PR #2, opened 2022-07-14
Add the PMC-Mean model type

skejserjensen:


server/src/models/pmcmean.rs
```diff
@@ -13,46 +13,131 @@
...
 pub fn grid(
     gid: i32,
     start_time: i64,
     end_time: i64,
     sampling_interval: i32,
     model: &[u8],
-    gaps: &[u8],
+    _gaps: &[u8],
     tids: &mut Int32Builder,
     timestamps: &mut TimestampMillisecondBuilder,
```
CGodiksen, line 137: Since the MQTT simulator generates timestamps with microsecond precision, we currently use a builder with TimestampMicrosecondType in the storage engine.

To avoid further confusion we should probably define type alias' in main that specify the type for single values, the arrays and the builders.

server/src/models/pmcmean.rs
```diff
@@ -13,46 +13,131 @@
...
+    _gaps: &[u8],
 ) -> f32 {
     let length = ((end_time - start_time) / sampling_interval as i64) + 1;
-    length as f32 * decode(model)
+    length as f32 * decode_model(model)
 }
 
+/// Reconstruct the data points for a time series segment whose values are
+///  represented by a model of type PMC-Mean.
 pub fn grid(
```
CGodiksen, line 129: Maybe make the effect of the function clearer by specifying that the data is reconstructed in the builders.

server/src/models/pmcmean.rs
```diff
@@ -61,7 +146,234 @@ pub fn grid(
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::num::f32;
+    use proptest::{prop_assert, prop_assume, proptest};
+
+    // Tests for PMCMeanModelType.
+    proptest! {
```
CGodiksen, line 168: Maybe indent tests within a "proptest" scope. If this is changed here it also needs to be changed under.

CGodiksen (changes requested):
Added a few comments. Another minor thing is the Copyright year being "2021". This might also need to be updated.

server/src/models/pmcmean.rs
```diff
@@ -13,46 +13,131 @@
...
+//!
+//! [Poor Man’s Compression paper]: https://ieeexplore.ieee.org/document/1260811
+//! [ModelarDB paper]: https://dl.acm.org/doi/abs/10.14778/3236187.3236215
+
+use std::error::Error;
 
 use datafusion::arrow::array::{Float32Builder, Int32Builder, TimestampMillisecondBuilder};
 
-/** Public Functions **/
+///The state the PMC-Mean model type needs while fitting a model to a time
```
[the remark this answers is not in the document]
skejserjensen: Fixed

server/src/models/pmcmean.rs
```diff
@@ -13,46 +13,131 @@
...
+/// series segment.
+struct PMCMeanModelType {
```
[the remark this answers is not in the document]
skejserjensen: I have renamed `PMCMeanModelType` to `PMCMean` as I think `PMCMeanModel` will cause confusion. PMC-Mean is a method for fitting a constant function (a model) to a time series segment within an error bound and thus not a model.

server/src/models/pmcmean.rs
skejserjensen: I have added the type aliases to a new module named [types.rs](https://github.com/ModelarData/MiniModelarDB/blob/e888c7f608b1608d7f0c50425477feb69cf323d9/server/src/types.rs) so we have a central place to define types used throughout the system. For the moment the types are defined based on the requirements of [pmcmean.rs](https://github.com/ModelarData/MiniModelarDB/blob/e888c7f608b1608d7f0c50425477feb69cf323d9/server/src/models/pmcmean.rs) as changing its types also requires changes to [tables.rs](https://github.com/ModelarData/MiniModelarDB/blob/e888c7f608b1608d7f0c50425477feb69cf323d9/server/src/tables.rs).

server/src/models/pmcmean.rs
skejserjensen: Fixed

server/src/models/pmcmean.rs
skejserjensen: Unfortunately `rustfmt` currently cannot format code inside macros #rust-lang/rustfmt/issues/8 and `proptest` does not provide other means for defining property-based tests #AltSysrq/proptest/issues/149. So, although I agree that the current formatting is less than optimal, I propose we keep it as is to not spend time fighting `rustfmt`.

skejserjensen:
> Added a few comments. Another minor thing is the Copyright year being "2021". This might also need to be updated.

I have updated the pull-request based on your suggestions. However, I do not think it is necessary to update the license header unless the entity that owns the copyright changes.

server/src/models/pmcmean.rs
CGodiksen: Small grammatical change. "appended" instead of "append".

server/src/types.rs
```diff
@@ -0,0 +1,50 @@
...
+pub type ArrowTimestamp = datafusion::arrow::datatypes::TimestampMillisecondType;
+
+// Types used for a collection of timestamps.
+pub type TimestampBuilder = datafusion::arrow::array::PrimitiveBuilder<ArrowTimestamp>;
+pub type TimestampArray = datafusion::arrow::array::PrimitiveArray<ArrowTimestamp>;
+
+// Types used for a single value.
+pub type Value = std::primitive::f32;
+pub type ArrowValue = datafusion::arrow::datatypes::Float32Type;
+#[cfg(test)] // Proptest is a dev-dependencies.
```
CGodiksen, line 42: Small grammatical change. "dev-dependency" instead of "dev-dependencies".

server/src/types.rs
```diff
@@ -0,0 +1,50 @@
...
+pub mod tests {
+    // proptest::num::f32 is not type.
```
CGodiksen, line 44: Maybe "is not a type" instead.

CGodiksen (changes requested):

server/src/models/pmcmean.rs
skejserjensen: Fixed

server/src/types.rs
skejserjensen: Fixed

server/src/types.rs
skejserjensen: Fixed

CGodiksen (approved):

agneborn98 (approved):

PR #6, opened 2022-07-20
Ingestion of MQTT messages.

CGodiksen:
This PR includes an Ingestor component that can connect to an MQTT broker, subscribe to specific topics and insert messages from these topics into the storage engine. It should be noted that this PR does not include unit tests since testing the ingestor requires an MQTT broker and simulator. This component should instead be tested as part of integration testing.

### DIFF
--- a/server/src/ingestion.rs
+++ b/server/src/ingestion.rs
@@ -36,12 +36,7 @@ pub struct Ingestor {
 
 impl Ingestor {
     /// Create a broker client, subscribe to the topics and start ingesting messages.
-    ///
-    /// # Arguments
-    /// * `compress_callback` - Function called on the messages when a batch of messages has been
-    /// ingested into memory. Due to possible memory limitations, the messages can also be supplied
-    /// to the compress callback through a file path.
-    pub fn start(self, compress_callback: fn(msg: Message)) {
+    pub fn start(self) {
         println!("Creating MQTT broker client.");
         let mut client = create_client(self.broker, self.client_id);
 
@@ -49,7 +44,7 @@ impl Ingestor {
             let mut stream = subscribe_to_broker(&mut client, self.topics, self.qos).await;
 
             println!("Waiting for messages...");
-            ingest_messages(&mut stream, &mut client, compress_callback).await;
+            ingest_messages(&mut stream, &mut client).await;
 
             Ok::<(), mqtt::Error>(())
         }) {
@@ -104,15 +99,11 @@ async fn subscribe_to_broker(
 
 // TODO: Send the messages to the storage engine when they are retrieved.
 /// Ingest the published messages in a loop until connection is lost.
-async fn ingest_messages(
-    stream: &mut AsyncReceiver<Option<Message>>,
-    client: &mut AsyncClient,
-    compress_callback: fn(msg: Message),
-) {
+async fn ingest_messages(stream: &mut AsyncReceiver<Option<Message>>, client: &mut AsyncClient) {
     // While the message stream resolves to the next item in the stream, ingest the messages.
     while let Some(msg_opt) = stream.next().await {
         if let Some(msg) = msg_opt {
-            compress_callback(msg)
+            println!("{}", msg);
         } else {
             // A "None" means we were disconnected. Try to reconnect...
             println!("Lost connection. Attempting reconnect.");
--- a/server/src/ingestion.rs
+++ b/server/src/ingestion.rs
@@ -41,7 +41,7 @@ impl Ingestor {
     /// Create a broker client, subscribe to the topics and start ingesting messages.
     pub fn start(self) {
         println!("Creating MQTT broker client.");
-        let mut client = create_client(self.broker, self.client_id);
+        let mut client = self.create_client();
 
         if let Err(err) = block_on(async {
             let mut stream = subscribe_to_broker(&mut client, self.topics, self.qos).await;
@@ -54,21 +54,19 @@ impl Ingestor {
             eprintln!("{}", err);
         }
     }
-}
 
-/// Create a broker client with the given broker URI and client ID.
-fn create_client(broker: &str, client_id: &str) -> AsyncClient {
-    let create_options = mqtt::CreateOptionsBuilder::new()
-        .server_uri(broker)
-        .client_id(client_id)
-        .finalize();
+    /// Create a broker client with the given broker URI and client ID.
+    fn create_client(&self) -> AsyncClient {
+        let create_options = mqtt::CreateOptionsBuilder::new()
+            .server_uri(self.broker)
+            .client_id(self.client_id)
+            .finalize();
 
-    let mut client = mqtt::AsyncClient::new(create_options).unwrap_or_else(|e| {
-        eprintln!("Error creating the client: {:?}", e);
-        process::exit(1);
-    });
-
-    client
+        mqtt::AsyncClient::new(create_options).unwrap_or_else(|e| {
+            eprintln!("Error creating the client: {:?}", e);
+            process::exit(1);
+        })
+    }
 }
 
 /// Make the connection to the broker and subscribe to the specified topics.
--- a/server/src/ingestion.rs
+++ b/server/src/ingestion.rs
@@ -15,9 +15,9 @@
 
 //! Support for data ingestion from an MQTT broker.
 //!
-//! To use the ingestor, run the "ingestor.start()" method. This method first creates the client.
-//! Then it uses the created client to connect to the broker and subscribe to the specified topics.
-//! The connection message stream is then looped over to ingest the messages that are published to the topics.
+//! To use the ingestor, run the "ingestor.start()" method. This method first creates the client. Then
+//! it uses the created client to connect to the broker and subscribe to the specified topics. The
+//! connection message stream is then looped over to ingest the messages that are published to the topics.
 
 use std::{process, time::Duration};
 
@@ -25,8 +25,10 @@ use futures::{executor::block_on, stream::StreamExt};
 use paho_mqtt as mqtt;
 use paho_mqtt::{AsyncClient, AsyncReceiver, Message};
 
-/// A single MQTT client that can subscribe to the specified broker and ingest messages from the
-/// specified topics. Note that after creation, the ingestor needs to be started to ingest messages.
+use crate::storage::StorageEngine;
+
+/// A single MQTT client that can subscribe to `broker` and ingest messages from `topics`. Note that
+/// after creation, the ingestor needs to be started to ingest messages.
 pub struct Ingestor {
     /// Server URI for the MQTT broker.
     broker: &'static str,
@@ -34,12 +36,17 @@ pub struct Ingestor {
     client_id: &'static str,
     /// Specific topics that should be subscribed to. Use "\[*]" to subscribe to all topics.
     topics: &'static [&'static str],
-    /// The quality of service for each subscribed-to topic. Should be of same length as `topics` field.
+    /// The quality of service for each subscribed-to topic. Should be same length as `topics` field.
     qos: &'static [i32],
 }
 
 impl Ingestor {
-    pub fn new(broker: &str, client_id: &str, topics: &[&str], qos: &[i32]) -> Self {
+    pub fn new(
+        broker: &'static str,
+        client_id: &'static str,
+        topics: &'static [&'static str],
+        qos: &'static [i32],
+    ) -> Self {
         Self {
             broker,
             client_id,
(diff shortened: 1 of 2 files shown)


agneborn98 (approved):
No comment

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! Support for data ingestion from an MQTT broker.
+//!
+//! To use the ingestor, run the "ingestor.start()" method. This method first creates the client. Then
+//! it uses the created client to connect to the broker and subscribe to the specified topics. The
+//! connection message stream is then looped over to ingest the messages that are published to the topics.
```
skejserjensen, line 20: Maybe "connection's"

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! Support for data ingestion from an MQTT broker.
+//!
+//! To use the ingestor, run the "ingestor.start()" method. This method first creates the client. Then
```
skejserjensen, line 18: Maybe `ingestor.start()` instead of "ingestor.start()".

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+use futures::{executor::block_on, stream::StreamExt};
+use paho_mqtt as mqtt;
+use paho_mqtt::{AsyncClient, AsyncReceiver, Message};
+
+use crate::storage::StorageEngine;
+
+/// A single MQTT client that can subscribe to `broker` and ingest messages from `topics`. Note that
+/// after creation, the ingestor needs to be started to ingest messages.
+pub struct Ingestor {
+    /// Server URI for the MQTT broker.
```
skejserjensen, line 33: Maybe remove "Server".

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+
+use futures::{executor::block_on, stream::StreamExt};
+use paho_mqtt as mqtt;
+use paho_mqtt::{AsyncClient, AsyncReceiver, Message};
+
+use crate::storage::StorageEngine;
+
+/// A single MQTT client that can subscribe to `broker` and ingest messages from `topics`. Note that
+/// after creation, the ingestor needs to be started to ingest messages.
+pub struct Ingestor {
```
skejserjensen, line 32: Unclear why all of the struct's field and method parameters have the lifetime [`'static`](https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html) instead of defining a [lifetime for the `Ingestor` struct.](https://doc.rust-lang.org/rust-by-example/scope/lifetime/struct.html)

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Server URI for the MQTT broker.
+    broker: &'static str,
+    /// ID that is used to uniquely identify the ingestor as a client to the MQTT broker.
+    client_id: &'static str,
+    /// Specific topics that should be subscribed to. Use "\[*]" to subscribe to all topics.
```
skejserjensen, line 37: Leftover \ ?

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+    topics: &'static [&'static str],
+    /// The quality of service for each subscribed-to topic. Should be same length as `topics` field.
```
skejserjensen, line 39: Maybe move the second part of the doc comment to `new()`, as that is the time when it is important to know `topics` and `qos` must be of the same length. Also, maybe remove "field" or change "`topics` field" to "the `topics` field".

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+    qos: &'static [i32],
+}
+
+impl Ingestor {
+    pub fn new(
```
skejserjensen, line 44: Maybe redefine the method to `try_new(...) -> Result<Self, Error or String>` where `Self` is only returned if `topics` and `qos` is the same size, and if possible, the `paho_mqtt` client can connect.

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+            broker,
+            client_id,
+            topics,
+            qos,
+        }
+    }
+
+    /// Create a broker client, subscribe to the topics, and start ingesting messages.
+    pub fn start(self, mut storage_engine: StorageEngine) {
+        println!("Creating MQTT broker client.");
```
skejserjensen, line 60: `println!` should be replaced with log message now that `tracing` have been added.

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+        let mut client = self.create_client();
+
+        if let Err(err) = block_on(async {
```
skejserjensen, line 63: Maybe immediately return the `Result` from `block_on()` using the [question mark operator](https://doc.rust-lang.org/std/result/index.html#the-question-mark-operator-) so the caller of `start()` can decide what to do? For example, log the error, restart ingestion, ignore that broker, etc.

server/src/ingestion.rs
skejserjensen, line 60: Maybe include `client_id` and `broker` in log messages when suitable.

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+            process::exit(1);
+        })
+    }
+
+    /// Make the connection to the broker and subscribe to the specified topics.
+    async fn subscribe_to_broker(
+        &self,
+        client: &mut AsyncClient,
+    ) -> AsyncReceiver<Option<Message>> {
+        // Get message stream before connecting since messages can arrive as soon as the connection is made.
```
skejserjensen, line 93: Maybe replace "Get message stream" with "Get the message stream" and add a comment that documents why 25 was chosen for `buffer_sz`.

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+        let mut stream = client.get_stream(25);
+
+        // Define last will and testament message to notify other clients about disconnect.
```
skejserjensen, line 96: Maybe replace "Define last will" with "Define the last will".

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+        let lwt = mqtt::Message::new("mdb_lwt", "ModelarDB lost connection", mqtt::QOS_1);
+
+        let connect_options = mqtt::ConnectOptionsBuilder::new()
+            .keep_alive_interval(Duration::from_secs(30))
```
skejserjensen, line 100: Maybe add a comment that documents why 30 seconds was chosen for the `keep_alive_interval`.

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+        println!("Connecting to MQTT broker.");
+        client.connect(connect_options).await;
+
+        println!("Subscribing to topics: {:?}", self.topics);
+        client.subscribe_many(self.topics, self.qos).await;
+
+        stream
+    }
+
+    /// Ingest the published messages in a loop until connection is lost.
```
skejserjensen, line 115: Maybe replace "until connection is lost" with "until the connection to the MQTT broker is lost".

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+    async fn ingest_messages(
+        stream: &mut AsyncReceiver<Option<Message>>,
+        client: &mut AsyncClient,
+        storage_engine: &mut StorageEngine,
+    ) {
+        // While the message stream resolves to the next item in the stream, ingest the messages.
```
skejserjensen, line 121: Maybe replace "resolves to" with "returns" and "item" with "message".

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+        while let Some(msg_opt) = stream.next().await {
+            if let Some(msg) = msg_opt {
+                storage_engine.insert_message(msg);
+            } else {
+                // A "None" means we were disconnected. Try to reconnect.
+                println!("Lost connection. Attempting reconnect.");
+                while let Err(err) = client.reconnect().await {
```
skejserjensen, line 128: It is unclear how this loop ever terminates if `client` is never able to reconnect.

server/src/ingestion.rs
```diff
@@ -0,0 +1,135 @@
...
+                    eprintln!("Error reconnecting: {:?}", err);
+                    tokio::time::sleep(Duration::from_millis(1000)).await;
```
skejserjensen, line 130: Maybe add a comment that documents why 1000ms was chosen as the duration to wait between reconnection attempts.

skejserjensen (changes requested):

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: Simplified the lifetime of the struct fields by changing to types to `String` and `Vec`.

server/src/ingestion.rs
CGodiksen: Since "" was used, the square brackets were removed by the comment formatter if not using a backslash. I have changed it to `[*]` instead.

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: I think it is a good idea but i was not able to get it to work since the return from `block_on` does not seem to be a simple `Result`. It might not be necessary to spend time on this since this component will be replaced in the near future.

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: 25 was only chosen because that was the default value used in the paho_mqtt example. The documentation or source does not give any information on how to best set this argument so I thought sticking with their default might be best. Should we mention this in the comment?

server/src/ingestion.rs
CGodiksen: Fixed-

server/src/ingestion.rs
CGodiksen: As with the `buffer_sz` this was chosen since it was the default in the paho_mqtt example and i did not found any reasoning to change it.  Should we mention this in the comment?

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: As with both `buffer_size` and `keep_alive_internal` this was chosen due to it being the standard. Should this be mentioned in a comment?

server/src/ingestion.rs
CGodiksen: Removed the reconnect feature. As mentioned earlier, it might no be relevant to spend time on more "advanced" functionality such as this since this component will be replaced in the near future

server/src/ingestion.rs
CGodiksen: Changed to `try_new` where `Self` is only returned if `topics` and `qos` is of the same length. Also added some tests for this.

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: This has been removed.

server/src/ingestion.rs
skejserjensen: Ok

server/src/ingestion.rs
skejserjensen: I agree.

server/src/ingestion.rs
skejserjensen: In general I think it is a good idea to mention were magic numbers come from if possible. So I think we should add a comment stating that it is from the paho_mqtt example.

skejserjensen (changes requested):

server/src/ingestion.rs
CGodiksen: Fixed.

server/src/ingestion.rs
CGodiksen: Fixed.

skejserjensen (approved):

PR #8, opened 2022-07-21
Spilling uncompressed data to Parquet files.

CGodiksen:
This PR includes a new feature for the storage engine, designed to handle possible memory issues. The amount of memory given to the storage engine is now defined with a constant and if we attempt to use more than the reserved amount, the uncompressed data is spilled to a Parquet file buffer. The buffering is designed to still allow external components, such as the compressor, to use the uncompressed data with the same interface, no matter if the data is in memory or in a file.

The current implementation has the limitation that we need to know how many sensors the system has before starting, since we need to calculate how much memory is left for storing uncompressed finished segments. This limitation is going to be removed in a future branch.

It should be noted that only the last 20 commits are new. The rest are commits from the original storage engine branch that this branch was created on.


server/src/storage/segment.rs
```diff
@@ -104,58 +116,109 @@ impl SegmentBuilder {
...
+
+        let expected = (segment_builder.timestamps.capacity() * mem::size_of::<Timestamp>())
+            + (segment_builder.values.capacity() * mem::size_of::<Value>());
+
+        assert_eq!(SegmentBuilder::get_memory_size(), expected)
+    }
+
+    #[test]
+    fn test_get_length() {
+        let data_point = get_data_point();
```
skejserjensen, line 202: The purpose of `data_point` is unclear in this test.

server/src/storage/segment.rs
```diff
@@ -165,15 +228,18 @@ mod tests {
     }
```
skejserjensen, line 228: Is this test missing an assert? If not a comment should probably be added with information about what the test checks.

server/src/storage/segment.rs
```diff
@@ -165,15 +228,18 @@ mod tests {
...
 
     #[test]
-    fn test_can_check_segment_is_not_full() {
-        let (data_point, mut segment_builder) = get_empty_segment_builder();
+    fn test_check_segment_is_not_full() {
+        let data_point = get_data_point();
```
skejserjensen, line 232: The purpose of `data_point` in this test is not clear.

server/src/storage/mod.rs
```diff
@@ -13,38 +13,52 @@
...
-
-const INITIAL_BUILDER_CAPACITY: usize = 50;
+// Note that the initial capacity has to be a multiple of 64 bytes to avoid the actual capacity
+// being larger due to internal alignment when allocating memory for the builders.
+const INITIAL_BUILDER_CAPACITY: usize = 64;
+// TODO: The sensor count should be dynamic and not predefined.
+const SENSOR_COUNT: usize = 2;
+const RESERVED_BYTES: usize = 5000;
 
 /// Keeping track of all uncompressed data, both while being built and when finished.
```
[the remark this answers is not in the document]
CGodiksen: Changed to "Manages all uncompressed..."

server/src/storage/mod.rs
```diff
@@ -82,11 +96,66 @@ impl StorageEngine {
...
+    /// Write `batch` to a persistent Apache Parquet file on disk.
+    pub fn save_compressed_data(key: String, first_timestamp: Timestamp, batch: RecordBatch) {
+        let folder_path = format!("compressed/{}", key);
+        fs::create_dir_all(&folder_path);
+
+        let path = format!("{}/{}.parquet", folder_path, first_timestamp);
+        write_batch_to_parquet(batch, path);
+    }
+
+    /// Move `segment_builder` to the the compression queue. If necessary, buffer the data first.
```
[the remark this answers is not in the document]
CGodiksen: I agree that buffering should only be used when referencing the in-memory segment. I have renamed everything related to the Parquet file segments to "spilled".

server/src/storage/mod.rs
```diff
@@ -82,11 +96,66 @@ impl StorageEngine {
...
+    fn queue_segment(&mut self, key: String, segment_builder: SegmentBuilder) {
```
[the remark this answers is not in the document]
CGodiksen: Changed to "enqueue_segment".

server/src/storage/mod.rs
```diff
@@ -82,11 +96,66 @@ impl StorageEngine {
...
+        println!("Saving the finished segment. Remaining bytes: {}", self.remaining_bytes);
+
+        let uncompressed_segment: Box<dyn UncompressedSegment>;
+        let builder_size = SegmentBuilder::get_memory_size();
+
+        // If there is not enough space for the finished segment, spill the data to a Parquet file.
+        if builder_size > self.remaining_bytes {
+            println!("Not enough memory for the finished segment. Buffering the segment.");
+
+            let buffered = BufferedSegment::new(key.clone(), segment_builder);
```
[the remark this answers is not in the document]
CGodiksen: So far I have seen them as negligible. It has been assumed that the reserved bytes would be dedicated purely to the actual data. I do not think this should change now but it is something that should be kept in mind when during further development, especially when considering a production workload.

server/src/storage/segment.rs
```diff
@@ -13,61 +13,61 @@
  * limitations under the License.
  */
 
-//! Provides support for inserting and storing data in an in-memory segment. Furthermore, the data
-//! can be retrieved as a structured record batch.
+//! Support for different kinds of uncompressed segments.
+//!
```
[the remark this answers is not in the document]
CGodiksen: It was mainly used to separate the "header" and "body" of the module description. I have removed the separation.

server/src/storage/segment.rs
```diff
@@ -13,61 +13,61 @@
...
+//! The SegmentBuilder struct provides support for inserting and storing data in an in-memory segment.
```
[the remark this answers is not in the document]
CGodiksen: I have renamed "BufferedSegment" to "SpilledSegment". Regarding the removal of `FinishedSegment`, I think I might change `FinishedSegment` in the next branch anyway so i will consider this further then.

server/src/storage/segment.rs
CGodiksen: There was no purpose, it has been removed.

server/src/storage/segment.rs
CGodiksen: There was no purpose, it has been removed.

server/src/storage/segment.rs
CGodiksen: I am not quite sure what you mean, the tests checks that the `is_full` function returns `true` when the segment is full.

skejserjensen: The comment was because I misunderstood `test_check_segment_is_full()` due to GitHub's interface, it implementation makes sense.

skejserjensen (approved):

agneborn98 (approved):

PR #13, opened 2022-07-27
Remove constant sensor count requirement.

CGodiksen:
The previous version of the storage engine had the requirement that the amount of sensors outputting data into the system had to be constant and known before starting. This requirement has been removed so it is now possible to add and remove sensors while the system is running.

### DIFF
--- a/server/src/storage/mod.rs
+++ b/server/src/storage/mod.rs
@@ -165,14 +165,13 @@ impl StorageEngine {
 
         // Iterate through the finished segments to find a segment that is in memory.
         for finished in self.compression_queue.iter_mut() {
-            if finished.uncompressed_segment.get_memory_size() > 0 {
-                info!("Spilling the segment with key '{}' to a Parquet file.", finished.key);
+            info!("Spilling the segment with key '{}' to a Parquet file.", finished.key);
 
-                finished.spill_segment();
-
+            if let Ok(path) = finished.spill_to_parquet() {
                 // Add the size of the segment back to the remaining reserved bytes.
                 self.remaining_bytes += SegmentBuilder::get_memory_size();
 
+                info!("Spilled the segment to '{}'. Remaining bytes: {}.", path, self.remaining_bytes);
                 return ();
             }
         }
--- a/server/src/storage/mod.rs
+++ b/server/src/storage/mod.rs
@@ -30,6 +30,7 @@ use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::parquet::arrow::ArrowWriter;
 use datafusion::parquet::basic::Encoding;
 use datafusion::parquet::file::properties::WriterProperties;
+use tracing::{error, info};
 
 use crate::storage::data_point::DataPoint;
 use crate::storage::segment::{SpilledSegment, FinishedSegment, SegmentBuilder, UncompressedSegment};
@@ -41,6 +42,31 @@ const INITIAL_BUILDER_CAPACITY: usize = 64;
 // TODO: The sensor count should be dynamic and not predefined.
 const SENSOR_COUNT: usize = 2;
 const RESERVED_BYTES: usize = 5000;
+
+// TODO: Maybe remove finished segment and add get_key to uncompressed segment.
+// TODO: Remove the sensor count constant.
+// TODO: Set the initial remaining bytes to be equal to reserved bytes.
+
+// TODO: Before creating a new builder, check if there is enough space for the builder.
+// TODO: If there is, create it and remove the builder size from the remaining bytes.
+// TODO: If there is not, we assume that it is because there is finished uncompressed segments in the queue.
+
+// TODO: We should find the first uncompressed, unbuffered finished segment in the queue and buffer it.
+// TODO: This could be done with a function on the UncompressedSegment trait.
+// TODO: It should return Ok if a unbuffered segment was buffered and Err if trying to buffer an already buffered segment.
+// TODO: If not able to find any unbuffered finished segments, we should panic.
+
+// TODO: When queueing a finished segment, add the builder size to the remaining bytes.
+
+// TODO: Add test for decrementing the remaining bytes when creating a builder.
+// TODO: Add test for buffering unbuffered finished segments if there is not enough space when creating a builder.
+// TODO: Add test for panicking if trying to buffer unbuffered and there are none.
+// TODO: Add test for incrementing the remaining bytes when a builder is finished.
+
+// TODO: Add test for checking that we cannot buffer a buffered finished segment.
+// TODO: Add test for checking that we can buffer a unbuffered finished segment.
+
+// TODO: Maybe split insert message into separate functions to avoid one large function.
 
 /// Manages all uncompressed data, both while being built and when finished.
 pub struct StorageEngine {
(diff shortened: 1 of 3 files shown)


server/src/storage/mod.rs
```diff
@@ -68,32 +67,37 @@ impl StorageEngine {
...
+                    // If there is not enough space for a new segment, spill a finished segment.
+                    if SegmentBuilder::get_memory_size() > self.remaining_bytes {
+                        self.spill_finished_segment();
+                    }
 
+                    // Create a new segment and remove the size from the reserved remaining memory.
                     let mut segment = SegmentBuilder::new();
-                    segment.insert_data(&data_point);
+                    self.remaining_bytes -= SegmentBuilder::get_memory_size();
+                    info!("Created segment. Remaining bytes: {}.", self.remaining_bytes);
```
skejserjensen, line 94: Maybe add `key` or use a `tracing::span` to make it clear for which time series the segment was created.

server/src/storage/segment.rs
```diff
@@ -41,6 +41,8 @@ pub trait UncompressedSegment {
     fn get_record_batch(&mut self) -> RecordBatch;
 
     fn get_memory_size(&self) -> usize;
+
+    fn spill_to_parquet(&mut self, key: String) -> Result<SpilledSegment, String>;
```
skejserjensen, line 45: Maybe add a comment documenting why `spill_to_parquet` is defined on `UncompressedSegment` instead of `SegmentBuilder`. Also, maybe return [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html) instead of `String`.

server/src/storage/segment.rs
```diff
@@ -123,6 +125,12 @@ impl UncompressedSegment for SegmentBuilder {
     fn get_memory_size(&self) -> usize {
         SegmentBuilder::get_memory_size()
     }
+
+    /// Spill the in-memory segment to a Parquet file and return Ok when finished.
+    fn spill_to_parquet(&mut self, key: String) -> Result<SpilledSegment, String> {
+        let batch = self.get_record_batch();
+        Ok(SpilledSegment::new(key.clone(), batch))
```
skejserjensen, line 132: Will the many `unwrap()` calls be removed soon or is the `StorageEngine` intentionally designed to `pacnic!()` without attempting recovery when an error occur?

Generally, I do not think `unwrap()` should be used to handle operations that can fail, e.g., `write_batch_to_parquet()` and `SpilledSegment::new()`, unless documented invariants ensure that the error cannot occur. Since writing files to disk can [fail](https://danluu.com/file-consistency/) in [many](https://danluu.com/filesystem-errors/) [interesting](https://danluu.com/deconstruct-files/) [ways](https://wiki.postgresql.org/wiki/Fsync_Errors), I think the `StorageEngine` generally should return errors and allow the caller to decide what to do instead of calling `panic!()` through `unwrap()`.

server/src/storage/segment.rs
```diff
@@ -132,18 +140,16 @@ pub struct SpilledSegment {
 }
 
 impl SpilledSegment {
-    /// Retrieve the data from the segment builder, save it to a Parquet file, and save the path.
-    pub fn new(key: String, mut segment_builder: SegmentBuilder) -> Self {
+    /// Save the data in `batch` to a Parquet file, and return a spilled segment with the path.
```
skejserjensen, line 143: Maybe replace "Save" with "Spil" to be consistent with `SpilledSegment` and `spill_to_parquet()`.

skejserjensen (changes requested):

server/src/storage/mod.rs
```diff
@@ -68,32 +67,37 @@ impl StorageEngine {
...
                         self.enqueue_segment(key, full_segment)
                     }
                 } else {
-                    println!(
-                        "Could not find segment with key '{}'. Creating segment.",
-                        key
-                    );
+                    info!("Could not find segment with key '{}'. Creating segment.", key);
+
+                    // If there is not enough space for a new segment, spill a finished segment.
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -68,32 +67,37 @@ impl StorageEngine {
...
+                    if SegmentBuilder::get_memory_size() > self.remaining_bytes {
+                        self.spill_finished_segment();
+                    }
 
+                    // Create a new segment and remove the size from the reserved remaining memory.
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -25,21 +25,20 @@ use std::collections::HashMap;
...
-use crate::storage::segment::{SpilledSegment, FinishedSegment, SegmentBuilder, UncompressedSegment};
+use crate::storage::segment::{FinishedSegment, SegmentBuilder};
 use crate::types::Timestamp;
 
 // Note that the initial capacity has to be a multiple of 64 bytes to avoid the actual capacity
 // being larger due to internal alignment when allocating memory for the builders.
 const INITIAL_BUILDER_CAPACITY: usize = 64;
-// TODO: The sensor count should be dynamic and not predefined.
-const SENSOR_COUNT: usize = 2;
 const RESERVED_BYTES: usize = 5000;
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -58,7 +57,7 @@ impl StorageEngine {
             // TODO: Maybe create with estimated capacity to avoid reallocation.
             data: HashMap::new(),
             compression_queue: VecDeque::new(),
-            remaining_bytes: RESERVED_BYTES - (SegmentBuilder::get_memory_size() * SENSOR_COUNT)
+            remaining_bytes: RESERVED_BYTES,
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: I agree that `unwrap()` should be used a lot less than it is right now. In general there needs to be more strict error handling in the storage engine so I have added attempting to remove `unwrap()` to that "feature".

server/src/storage/mod.rs
```diff
@@ -68,32 +67,37 @@ impl StorageEngine {
...
 
                 if let Some(segment) = self.data.get_mut(&key) {
-                    println!("Found existing segment with key '{}'.", key);
+                    info!("Found existing segment with key '{}'.", key);
 
                     segment.insert_data(&data_point);
 
                     if segment.is_full() {
-                        println!("Segment is full, moving it to the compression queue.");
+                        info!("Segment is full, moving it to the compression queue.");
```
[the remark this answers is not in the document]
CGodiksen: Using a span for the `insert_message` context that contains the key that is inserted into. Also removed all mentions of the key in individual log messages.

server/src/storage/mod.rs
CGodiksen: Fixed.

skejserjensen (approved):

PR #18, opened 2022-08-03
Saving compressed segments together. 

CGodiksen:
This PR includes an extension to the functionality of the storage engine to avoid unnecessary I/O and very small files. Previously each compressed segment was saved in a separate Parquet file but now the storage engine stores the compressed data in memory if possible and saves compressed data in batches when necessary.

### DIFF
--- a/server/src/storage/mod.rs
+++ b/server/src/storage/mod.rs
@@ -45,7 +45,7 @@ use crate::storage::time_series::CompressedTimeSeries;
 const INITIAL_BUILDER_CAPACITY: usize = 64;
 const UNCOMPRESSED_RESERVED_MEMORY_IN_BYTES: usize = 5000;
 const COMPRESSED_RESERVED_MEMORY_IN_BYTES: isize = 5000;
-const MAX_COMPRESSED_FILE_SIZE_IN_BYTES: usize = 10240;
+const MAX_COMPRESSED_FILE_SIZE_IN_BYTES: usize = 1024;
 
 /// Manages all uncompressed data, both while being built and when finished.
 pub struct StorageEngine {
@@ -107,7 +107,11 @@ impl StorageEngine {
             // Create a new segment and reduce the remaining amount of reserved memory by its size.
             let mut segment = SegmentBuilder::new();
             self.uncompressed_remaining_memory_in_bytes -= SegmentBuilder::get_memory_size();
-            info!("Created segment. Remaining bytes: {}.", self.uncompressed_remaining_memory_in_bytes);
+
+            info!(
+                "Created segment. Remaining bytes: {}.",
+                self.uncompressed_remaining_memory_in_bytes
+            );
 
             segment.insert_data(&data_point);
             self.uncompressed_data.insert(key, segment);
@@ -231,6 +235,8 @@ mod tests {
     use super::*;
     use std::time::{SystemTime, UNIX_EPOCH};
 
+    use crate::storage::time_series::test_util;
+
     // Tests for uncompressed data.
     #[test]
     fn test_cannot_insert_invalid_message() {
@@ -312,8 +318,7 @@ mod tests {
         let reserved_memory = storage_engine.uncompressed_remaining_memory_in_bytes;
 
         // If there is enough reserved memory to hold n builders, we need to create n + 1 to panic.
-        for i in 0..(reserved_memory / SegmentBuilder::get_memory_size()) + 1
-        {
+        for i in 0..(reserved_memory / SegmentBuilder::get_memory_size()) + 1 {
             insert_generated_message(&mut storage_engine, i.to_string());
         }
     }
@@ -349,22 +354,47 @@ mod tests {
 
     // Tests for compressed data.
     #[test]
-    fn test_cannot_insert_valid_compressed_segment() {
+    #[should_panic(expected = "Schema of record batch does not match compressed segment schema.")]
+    fn test_cannot_insert_invalid_compressed_segment() {
+        let invalid = test_util::get_invalid_compressed_segment_record_batch();
+        let mut storage_engine = StorageEngine::new();
 
+        storage_engine.insert_compressed_data("key".to_owned(), invalid);
     }
 
     #[test]
     fn test_can_insert_compressed_segment_into_new_time_series() {
+        let segment = test_util::get_compressed_segment_record_batch();
+        let mut storage_engine = StorageEngine::new();
 
+        storage_engine.insert_compressed_data("key".to_owned(), segment);
+
+        assert!(storage_engine.compressed_data.contains_key("key"));
+        assert_eq!(storage_engine.compression_queue.pop_front().unwrap(), "key");
+
+        assert!(storage_engine.compressed_data.get("key").unwrap().size_in_bytes > 0);
     }
 
     #[test]
     fn test_can_insert_compressed_segment_into_existing_time_series() {
+        let segment = test_util::get_compressed_segment_record_batch();
+        let mut storage_engine = StorageEngine::new();
 
+        storage_engine.insert_compressed_data("key".to_owned(), segment.clone());
+        let previous_size = storage_engine.compressed_data.get("key").unwrap().size_in_bytes;
+        storage_engine.insert_compressed_data("key".to_owned(), segment);
+
+        assert!(storage_engine.compressed_data.get("key").unwrap().size_in_bytes > previous_size);
     }
 
     #[test]
     fn test_remaining_bytes_decremented_when_inserting() {
+        let segment = test_util::get_compressed_segment_record_batch();
+        let mut storage_engine = StorageEngine::new();
+        let initial_remaining_memory = storage_engine.compressed_remaining_memory_in_bytes;
 
+        storage_engine.insert_compressed_data("key".to_owned(), segment);
+
+        assert!(storage_engine.compressed_remaining_memory_in_bytes < initial_remaining_memory);
     }
 }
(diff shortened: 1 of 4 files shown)


server/src/storage/segment.rs
```diff
@@ -142,7 +142,7 @@ pub struct SpilledSegment {
 impl SpilledSegment {
     /// Spill the data in `batch` to a Parquet file, and return a spilled segment with the path.
     pub fn new(key: String, batch: RecordBatch) -> Self {
-        let folder_path = format!("uncompressed/{}", key);
+        let folder_path = format!("storage/{}/uncompressed", key);
```
skejserjensen, line 145: Is "storage" a placeholder until the `StorageEngine` is integrated and the user-defined data folder can be used?

server/src/storage/time_series.rs
```diff
@@ -0,0 +1,212 @@
...
+            "Schema of record batch does not match compressed segment schema."
+        );
+
+        self.compressed_segments.push(segment);
+        self.size_in_bytes += segment_size;
+
+        segment_size
+    }
+
+    /// If the compressed segments are successfully saved to Parquet, return Ok, otherwise return Err.
```
skejserjensen, line 60: For consistency, should we include "Apache" when we reference an Apache project? At the moment it is inconsistent. Generally I think we should include it as part of the name.

server/src/storage/time_series.rs
```diff
@@ -0,0 +1,212 @@
...
+            // Combine the segments into a single record batch.
+            let schema = Self::get_compressed_segment_schema();
+            let batch = RecordBatch::concat(&Arc::new(schema), &*self.compressed_segments).unwrap();
+
+            // Create the folder structure if it does not already exist.
+            let folder_path = format!("storage/{}/compressed", key);
+            fs::create_dir_all(&folder_path)?;
+
+            // Create a path that uses the first timestamp as the filename.
+            let start_times: &TimestampArray = batch.column(2).as_any().downcast_ref().unwrap();
```
skejserjensen, line 77: Is the first timestamp used with indexing in mind (e.g., pruning data that is too new or old  to be part of the query) if so it should probably be documented?

server/src/storage/time_series.rs
```diff
@@ -0,0 +1,212 @@
...
+        let result = empty_time_series.save_to_parquet("key".to_owned());
+
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_get_size_of_segment() {
+        let segment = test_util::get_compressed_segment_record_batch();
+
+        assert_eq!(CompressedTimeSeries::get_size_of_segment(&segment), 2032);
```
skejserjensen, line 161: Maybe define a named constant in `test_util` with this information.

server/src/storage/time_series.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+#[cfg(test)]
+pub mod test_util {
```
skejserjensen, line 166: For consistency should test helper functions be in a separate module or the same module as the tests? Currently it looks like most modules define a single test module that includes both the tests and the helper functions.

server/src/storage/time_series.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    use datafusion::arrow::array::{BinaryArray, Float32Array, UInt8Array};
+    use datafusion::arrow::datatypes::DataType::UInt8;
+    use datafusion::arrow::datatypes::{Field, Schema};
+    use datafusion::arrow::record_batch::RecordBatch;
+
+    use crate::storage::time_series::CompressedTimeSeries;
+    use crate::types::{TimestampArray, ValueArray};
+
+    pub fn get_invalid_compressed_segment_record_batch() -> RecordBatch {
```
skejserjensen, line 177: Maybe document why the `RecordBatch` is invalid.

skejserjensen (changes requested):

server/src/storage/time_series.rs
agneborn98: I concur.

server/src/storage/mod.rs
```diff
@@ -13,101 +13,117 @@
...
-                    info!("Found existing segment.");
+            segment.insert_data(&data_point);
 
-                    segment.insert_data(&data_point);
+            if segment.is_full() {
+                info!("Segment is full, moving it to the queue of finished segments.");
 
-                    if segment.is_full() {
-                        info!("Segment is full, moving it to the compression queue.");
+                let full_segment = self.uncompressed_data.remove(&key).unwrap();
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -116,39 +132,61 @@ impl StorageEngine {
...
-    /// Write `batch` to a persistent Apache Parquet file on disk.
-    pub fn save_compressed_data(key: String, first_timestamp: Timestamp, batch: RecordBatch) {
-        let folder_path = format!("compressed/{}", key);
-        fs::create_dir_all(&folder_path);
+    /// Insert `segment` into the in-memory compressed time series buffer.
+    pub fn insert_compressed_data(&mut self, key: String, segment: RecordBatch) {
+        let _span = info_span!("insert_compressed_segment", key = key.clone()).entered();
+        info!("Inserting batch with {} rows into compressed time series.", segment.num_rows());
+
+        let compressed_segment_size;
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -13,101 +13,117 @@
...
+
+// TODO: Look into moving handling of uncompressed and compressed data into separate structs.
+// TODO: Look into custom errors for all errors in storage engine.
 
 // Note that the initial capacity has to be a multiple of 64 bytes to avoid the actual capacity
 // being larger due to internal alignment when allocating memory for the builders.
 const INITIAL_BUILDER_CAPACITY: usize = 64;
-const RESERVED_MEMORY_IN_BYTES: usize = 5000;
+const UNCOMPRESSED_RESERVED_MEMORY_IN_BYTES: usize = 5000;
+const COMPRESSED_RESERVED_MEMORY_IN_BYTES: isize = 5000;
```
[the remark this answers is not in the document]
CGodiksen: The main reason is that for uncompressed data we can check the remaining memory before allocating space since we first actually use the memory when creating a builder. For compressed data the memory is occupied as soon as the compressed segment is received so instead of checking the remaining memory before inserting, we instead insert it first and check if the remaining memory is below 0 after. 

This should be okay to do briefly since the memory is allocated for the compressed segment already anyway and can therefore be seen as being part of the memory overhead of running the storage engine. 

I have added a comment to the const that tries to explain this.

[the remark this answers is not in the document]
CGodiksen: The alternative would be to check the remaining bytes before inserting the compressed data which is also an option. This does make the structure of `insert_compressed_data` a bit more confusing. However, I think the structure needs to change when adding the `MAX_COMPRESSED_FILE_SIZE` setting later anyway, so i can look into this issue again then.

server/src/storage/mod.rs
```diff
@@ -157,6 +195,27 @@ impl StorageEngine {
         // If not able to find any in-memory finished segments, we should panic.
```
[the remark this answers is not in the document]
CGodiksen: I have added a TODO to save both uncompressed and compressed data in memory to disk before panicking.

server/src/storage/segment.rs
CGodiksen: Yes, I have added a TODO to change this after a user-defined data folder is given.

server/src/storage/time_series.rs
CGodiksen: I agree. Should we only change it in comments or should we also change functions such as `save_to_parquet` to `save_to_apache_parquet`? Currently, I have just changed it in the comments.

server/src/storage/time_series.rs
CGodiksen: Technically it was just used since it was unique but i have mentioned the thing about pruning in the comment.

server/src/storage/time_series.rs
CGodiksen: Added.

server/src/storage/time_series.rs
CGodiksen: I agree that helper functions should be in the same module as the tests normally. The only reason i used a separate module here is that these helper functions are used both here and in the `StorageEngine` tests. I have added a module doc comment to make this clear.

server/src/storage/time_series.rs
CGodiksen: Fixed.

server/src/storage/time_series.rs
skejserjensen: I suggest we change it for both comments and code to keep it consistent.

agneborn98 (approved):

server/src/storage/time_series.rs
agneborn98: Sounds good.

CGodiksen: I have changed it in the code. It should be noted that this means that `ArrowTimestamp` and `ArrowValue` should probably also be updated. There are also a few more places outside the storage engine where this needs to be changed, but I do not think this is the correct branch to do this in.

skejserjensen: I agree.

skejserjensen (approved):

PR #21, opened 2022-08-08
General Apache Parquet functionality

CGodiksen:
Since reading from, writing to, and identifying Apache Parquet files is needed multiple places in the system, it was decided to add this functionality in the storage engine. This PR includes three new functions in the storage engine that allows interaction with Apache Parquet files. The functions are tested using the [tempfile](https://crates.io/crates/tempfile) crate that can create a temporary directory which is deleted once out of scope. 

The places in the storage engine and catalog that used their own versions of this Apache Parquet functionality has been replaced by the shared functionality. 

It should also be noted that two basic functions to retrieve the schemas used for uncompressed and compressed data has been added to the storage engine.

### DIFF
--- a/server/src/storage/mod.rs
+++ b/server/src/storage/mod.rs
@@ -24,6 +24,8 @@ mod time_series;
 use std::collections::vec_deque::VecDeque;
 use std::collections::HashMap;
 use std::fs::File;
+use std::io::Read;
+use std::path::Path;
 use std::sync::Arc;
 
 use datafusion::arrow::datatypes::{ArrowPrimitiveType, DataType, Field, Schema};
@@ -52,6 +54,11 @@ use crate::types::{ArrowTimestamp, ArrowValue};
 const INITIAL_BUILDER_CAPACITY: usize = 64;
 const UNCOMPRESSED_RESERVED_MEMORY_IN_BYTES: usize = 5000;
 const COMPRESSED_RESERVED_MEMORY_IN_BYTES: isize = 5000;
+
+/// The expected [first four bytes of any Apache Parquet file].
+///
+/// [first four bytes of any Apache Parquet file]: https://en.wikipedia.org/wiki/List_of_file_signatures
+const APACHE_PARQUET_FILE_SIGNATURE: &[u8] = &[80, 65, 82, 49]; // PAR1.
 
 /// Manages all uncompressed and compressed data, both while being built and when finished.
 pub struct StorageEngine {
@@ -199,9 +206,9 @@ impl StorageEngine {
     /// file was written successfully, otherwise return `ParquetError`.
     pub fn write_batch_to_apache_parquet_file(
         batch: RecordBatch,
-        path: String
+        path: &Path
     ) -> Result<(), ParquetError> {
-        if let Ok(file) = File::create(path.clone()) {
+        if let Ok(file) = File::create(path) {
             let props = WriterProperties::builder()
                 .set_dictionary_enabled(false)
                 .set_encoding(Encoding::PLAIN)
@@ -214,17 +221,19 @@ impl StorageEngine {
             Ok(())
         } else {
             Err(ParquetError::General(
-                format!("Apache Parquet file at path '{}' could not be created.", path)
+                format!("Apache Parquet file at path '{}' could not be created.", path.display())
             ))
         }
     }
 
     /// Read all rows from the Apache Parquet file at the location given by `path` and return them
     /// in a record batch. If the file could not be read successfully, `ParquetError` is returned.
-    pub fn read_batch_from_apache_parquet_file(path: String) -> Result<RecordBatch, ParquetError> {
-        let error = format!("Apache Parquet file at path '{}' could not be read.", path);
+    pub fn read_batch_from_apache_parquet_file(path: &Path) -> Result<RecordBatch, ParquetError> {
+        let error = format!(
+            "Apache Parquet file at path '{}' could not be read.", path.display()
+        );
 
-        if let Ok(file) = File::open(&path) {
+        if let Ok(file) = File::open(path) {
             let reader = SerializedFileReader::new(file)?;
 
             // Extract the total row count from the file metadata.
@@ -246,6 +255,17 @@ impl StorageEngine {
             Ok(batch)
         } else {
             Err(ParquetError::General(error.to_owned()))
+        }
+    }
+
+    /// Return `true` if `path` is a readable Apache Parquet file, otherwise `false`.
+    fn is_path_an_apache_parquet_file(path: &Path) -> bool {
+        if let Ok(mut file) = File::open(path) {
+            let mut first_four_bytes = vec![0u8; 4];
+            file.read_exact(&mut first_four_bytes);
+            first_four_bytes == APACHE_PARQUET_FILE_SIGNATURE
+        } else {
+            false
         }
     }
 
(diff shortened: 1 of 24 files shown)


server/src/storage/mod.rs
```diff
@@ -167,6 +183,103 @@ impl StorageEngine {
...
+
+            // Extract the total row count from the file metadata.
+            let apache_parquet_metadata = reader.metadata();
+            let row_count = apache_parquet_metadata
+                .row_groups()
+                .iter()
+                .map(|rg| rg.num_rows())
+                .sum::<i64>() as usize;
+
+            // Read the data and convert it into a record batch.
```
skejserjensen, line 258: Maybe replace "into" with "to".

server/src/storage/mod.rs
```diff
@@ -403,4 +506,110 @@ mod tests {
...
+    #[test]
+    fn test_is_parquet_path_apache_parquet_file() {
+        let file_name = "test.parquet".to_owned();
+        let (_temp_dir, path, _batch) = create_file_in_temp_dir(file_name);
+
+        assert!(StorageEngine::is_path_an_apache_parquet_file(path.as_path()));
+    }
+
+    /// Create an Apache Parquet file from a generated record batch in the temp dir.
+    fn create_file_in_temp_dir(file_name: String) -> (TempDir, PathBuf, RecordBatch) {
```
skejserjensen, line 596: Maybe rename to `create_apache_parquet_file_in_temp_dir()`.

server/src/storage/segment.rs
```diff
@@ -128,14 +121,15 @@ impl UncompressedSegment for SegmentBuilder {
 
     /// Spill the in-memory segment to an Apache Parquet file and return Ok when finished.
     fn spill_to_apache_parquet(&mut self, key: String) -> Result<SpilledSegment, std::io::Error> {
```
skejserjensen, line 123: Maybe import `std::io::error` as `IOErrro` to differentiate it from `ParquetError`.

server/src/storage/segment.rs
```diff
@@ -128,14 +121,15 @@ impl UncompressedSegment for SegmentBuilder {
...
-        let batch = self.get_record_batch();
+        let batch = self.get_record_batch().unwrap();
```
skejserjensen, line 124: Maybe either convert to `IOError` and return with `?` or document with a comment why `unwrap()` is ok.

skejserjensen (changes requested):
One general comment/question. For consistency should we write subroutines that does not take a type instance as input (`self`) as methods on a type (e.g., `StorageEngine::get_compressed_segment_schema()`) or free functions in a module (e.g., `remote::start_arrow_flight_server()`)?

server/src/storage/mod.rs
```diff
@@ -167,6 +183,103 @@ impl StorageEngine {
...
+                Err(error)
+            }
+        } else {
+            Err(error)
+        }
+    }
+
+    /// Read all rows from the Apache Parquet file at the location given by `path` and return them
+    /// in a record batch. If the file could not be read successfully, `ParquetError` is returned.
+    pub fn read_batch_from_apache_parquet_file(path: &Path) -> Result<RecordBatch, ParquetError> {
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -167,6 +183,103 @@ impl StorageEngine {
...
+            Field::new("end_time", ArrowTimestamp::DATA_TYPE, false),
+            Field::new("values", DataType::Binary, false),
+            Field::new("min_value", ArrowValue::DATA_TYPE, false),
+            Field::new("max_value", ArrowValue::DATA_TYPE, false),
+            Field::new("error", DataType::Float32, false),
+        ])
+    }
+
+    // TODO: Test using more efficient encoding. Plain encoding makes it easier to read the files externally.
+    /// Write `batch` to an Apache Parquet file at the location given by `path`. Return Ok if the
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -167,6 +183,103 @@ impl StorageEngine {
...
+        batch: RecordBatch,
+        path: &Path,
+    ) -> Result<(), ParquetError> {
+        let error = ParquetError::General(
+            format!("Apache Parquet file at path '{}' could not be created.", path.display())
+        );
+
+        // Check if the extension of the given path is correct.
+        if path.extension().and_then(OsStr::to_str) == Some("parquet") {
+            if let Ok(file) = File::create(path) {
```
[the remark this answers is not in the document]
CGodiksen: I was actually looking into this briefly since I did not like the two `else` blocks right after each other and large indentation but i was not able to find a fix. `map_err()` looks good and i have replaced the `if else` with it.

server/src/storage/mod.rs
```diff
@@ -167,6 +183,103 @@ impl StorageEngine {
...
+    }
+
+    /// Read all rows from the Apache Parquet file at the location given by `path` and return them
+    /// in a record batch. If the file could not be read successfully, `ParquetError` is returned.
+    pub fn read_batch_from_apache_parquet_file(path: &Path) -> Result<RecordBatch, ParquetError> {
+        let error = ParquetError::General(
+            format!("Apache Parquet file at path '{}' could not be read.", path.display())
+        );
+
+        if let Ok(file) = File::open(path) {
```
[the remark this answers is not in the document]
CGodiksen: Fixed for consistency with `write_batch_to_apache_parquet_file` and to avoid the large indented block.

server/src/storage/mod.rs
CGodiksen: Fixed.

server/src/storage/mod.rs
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: Added comment explaining why `unwrap` is safe to use here.

CGodiksen:
> One general comment/question. For consistency should we write subroutines that does not take a type instance as input (`self`) as methods on a type (e.g., `StorageEngine::get_compressed_segment_schema()`) or free functions in a module (e.g., `remote::start_arrow_flight_server()`)?

I have fixed the comments. To answer your question I am not sure what would be best. I personally have no preference since I assume that it would change the function call to `storage::write_batch_to_apache_parquet_file` which also seems fine.

skejserjensen (approved):
> I have fixed the comments. To answer your question I am not sure what would be best. I personally have no preference since I assume that it would change the function call to `storage::write_batch_to_apache_parquet_file` which also seems fine.

I also do not have a strong preference either way, however, I prefer that the use of methods and free functions are consistent throughout the code. Maybe we should look into what generally is recommended in the Rust community, and if we cannot find anything, look at what is recommended for other languages that support both methods and functions like C++ and Python. The topic seems to be discussed a lot on Stack Overflow with regards to Python, e.g., see this recommendations to [create a static method, an instance method, or a function based on what it requires access to](https://stackoverflow.com/a/11788267). Of course this recommendation cannot be mapped directly to Rust as a `new` method definitely should be a static method on the type despite also working as a free function.

agneborn98 (approved):

PR #24, opened 2022-08-12
New storage engine structure.

CGodiksen:
This PR includes a significant change to the storage engine structure. Handling of uncompressed data is moved from the main `StorageEngine` struct into a new `UncompressedDataManager` struct. The same is done for handling of compressed data. Paths used in the storage engine has also been changed from `String` to `Path` and documentation has been changed to include links. 

Note that only the last 14 commits are new. Also note that most of the code has not been changed, only moved around.

### DIFF
--- a/server/src/storage/compressed_data_manager.rs
+++ b/server/src/storage/compressed_data_manager.rs
@@ -34,7 +34,7 @@ pub struct CompressedDataManager {
     data_folder_path: PathBuf,
     /// The compressed segments before they are saved to persistent storage.
     compressed_data: HashMap<String, CompressedTimeSeries>,
-    /// Prioritized queue of keys referring to [`CompressedTimeSeries`] that can be saved to persistent storage.
+    /// FIFO queue of keys referring to [`CompressedTimeSeries`] that can be saved to persistent storage.
     compressed_queue: VecDeque<String>,
     /// How many bytes of memory that are left for storing compressed segments.
     compressed_remaining_memory_in_bytes: isize,
@@ -52,14 +52,14 @@ impl CompressedDataManager {
     }
 
     /// Insert `segment` into the in-memory compressed time series buffer.
-    pub fn insert_compressed_data(&mut self, key: String, segment: RecordBatch) {
+    pub fn insert_compressed_segment(&mut self, key: String, segment: RecordBatch) {
         let _span = info_span!("insert_compressed_segment", key = key.clone()).entered();
         info!(
             "Inserting batch with {} rows into compressed time series.",
             segment.num_rows()
         );
 
-        // Since the compressed segment is already in memory, insert the segment in to the structure
+        // Since the compressed segment is already in memory, insert the segment into the structure
         // first and check if the reserved memory limit is exceeded after.
         let segment_size = if let Some(time_series) = self.compressed_data.get_mut(&key) {
             info!("Found existing compressed time series.");
@@ -87,7 +87,7 @@ impl CompressedDataManager {
 
     /// Save [`CompressedTimeSeries`] to disk until the reserved memory limit is no longer exceeded.
     fn save_compressed_data(&mut self) {
-        info!("Out of memory to store compressed data. Saving compressed data to disk.");
+        info!("Out of memory for compressed data. Saving compressed data to disk.");
 
         while self.compressed_remaining_memory_in_bytes < 0 {
             let key = self.compressed_queue.pop_front().unwrap();
@@ -124,7 +124,7 @@ mod tests {
         let invalid = test_util::get_invalid_compressed_segment_record_batch();
         let (_temp_dir, mut data_manager) = create_compressed_data_manager();
 
-        data_manager.insert_compressed_data("key".to_owned(), invalid);
+        data_manager.insert_compressed_segment("key".to_owned(), invalid);
     }
 
     #[test]
@@ -132,7 +132,7 @@ mod tests {
         let segment = test_util::get_compressed_segment_record_batch();
         let (_temp_dir, mut data_manager) = create_compressed_data_manager();
 
-        data_manager.insert_compressed_data("key".to_owned(), segment);
+        data_manager.insert_compressed_segment("key".to_owned(), segment);
 
         assert!(data_manager.compressed_data.contains_key("key"));
         assert_eq!(data_manager.compressed_queue.pop_front().unwrap(), "key");
@@ -144,9 +144,9 @@ mod tests {
         let segment = test_util::get_compressed_segment_record_batch();
         let (_temp_dir, mut data_manager) = create_compressed_data_manager();
 
-        data_manager.insert_compressed_data("key".to_owned(), segment.clone());
+        data_manager.insert_compressed_segment("key".to_owned(), segment.clone());
         let previous_size = data_manager.compressed_data.get("key").unwrap().size_in_bytes;
-        data_manager.insert_compressed_data("key".to_owned(), segment);
+        data_manager.insert_compressed_segment("key".to_owned(), segment);
 
         assert!(data_manager.compressed_data.get("key").unwrap().size_in_bytes > previous_size);
     }
@@ -160,7 +160,7 @@ mod tests {
         // Insert compressed data into the storage engine until data is saved to Apache Parquet.
         let max_compressed_segments = reserved_memory / test_util::COMPRESSED_SEGMENT_SIZE;
         for _ in 0..max_compressed_segments + 1 {
-            data_manager.insert_compressed_data("modelardb-test".to_owned(), segment.clone());
+            data_manager.insert_compressed_segment("modelardb-test".to_owned(), segment.clone());
         }
 
         // The compressed data should be saved to the "compressed" folder under the key.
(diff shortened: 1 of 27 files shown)


server/src/storage/segment.rs
```diff
@@ -33,20 +33,20 @@ use crate::storage::data_point::DataPoint;
...
-/// and spilled segments.
+/// Shared functionality between different types of uncompressed segments, such as [`SegmentBuilder`]
+/// and [`SpilledSegment`].
 pub trait UncompressedSegment {
     fn get_record_batch(&mut self) -> Result<RecordBatch, ParquetError>;
 
     fn get_memory_size(&self) -> usize;
 
-    // Since both segment builders and spilled segments are present in the compression queue, both
+    // Since both SegmentBuilders and SpilledSegments are present in the queue of finished segments, both
```
skejserjensen, line 43: Looks like links and `` are missing for `SegmentBuilder` and `SpilledSegment`. As you mentioned in a [comment](https://github.com/ModelarData/ModelarDB-RS/pull/23#discussion_r944129790) on #23 we can either use \[\`SegmentBuilder\`\]s or \[\`SegmentBuilders\`\]\(SegmentBuilder\). Generally when I have written documentation or papers I have only highlighted the name and not the plural s, but if you prefer that both are highlighted as you stated in the [comment](https://github.com/ModelarData/ModelarDB-RS/pull/23#discussion_r944129790) then that is fine by me.

server/src/storage/segment.rs
```diff
@@ -13,16 +13,16 @@
  * limitations under the License.
  */
 
-//! Support for different kinds of uncompressed segments. The SegmentBuilder struct provides support
-//! for inserting and storing data in an in-memory segment. SpilledSegment provides support for
+//! Support for different kinds of uncompressed segments. The [`SegmentBuilder`] struct provides support
```
skejserjensen, line 16: Maybe replace the instances of "uncompressed segment" with "UncompressedSegment" or "types implementing UncompressedSegment".

server/src/storage/segment.rs
skejserjensen, line 43: Maybe replace the instances of "finished segment" with "FinishedSegment".

server/src/storage/segment.rs
```diff
@@ -33,20 +33,20 @@ use crate::storage::data_point::DataPoint;
...
+    // Since both SegmentBuilders and SpilledSegments are present in the queue of finished segments, both
     // structs need to implement spilling to Apache Parquet, with already spilled segments returning Err.
-    fn spill_to_apache_parquet(&mut self, folder_path: String) -> Result<SpilledSegment, IOError>;
+    fn spill_to_apache_parquet(&mut self, folder_path: &Path) -> Result<SpilledSegment, IOError>;
 }
 
 /// A single segment being built, consisting of an ordered sequence of timestamps and values. Note
-/// that since array builders are used, the data can only be read once the builders are finished and
+/// that since [`ArrayBuilder`] is used, the data can only be read once the builders are finished and
 /// cannot be further appended to after.
```
skejserjensen, line 50: Maybe remove "after".

server/src/storage/uncompressed_data_manager.rs
```diff
@@ -0,0 +1,345 @@
...
+
+/// Converts raw MQTT messages to uncompressed data points and stores uncompressed data points
+/// temporarily in an in-memory buffer that spills to Apache Parquet files. When finished the data
+/// is made available for compression.
+pub struct UncompressedDataManager {
+    /// Path to the folder containing all uncompressed data managed by the [`StorageEngine`].
+    data_folder_path: PathBuf,
+    /// The uncompressed segments while they are being built.
+    uncompressed_data: HashMap<String, SegmentBuilder>,
+    /// Prioritized queue of finished segments that are ready for compression.
```
skejserjensen, line 37: Maybe add how the `FinishedSegment`s are prioritized.

server/src/storage/uncompressed_data_manager.rs
```diff
@@ -0,0 +1,345 @@
...
+        let finished_segment = FinishedSegment {
+            key,
+            uncompressed_segment: Box::new(segment_builder),
+        };
+
+        self.finished_queue.push_back(finished_segment);
+    }
+
+    /// Spill the first in-memory [`FinishedSegment`] in the queue of finished segments. If no
+    /// in-memory finished segments could be found, [`panic`](core::panic).
```
skejserjensen, line 124: As we also use [`std`](https://doc.rust-lang.org/std/index.html) and not just [`core`](https://doc.rust-lang.org/core/index.html) I suggest we link to `std` instead of `core`, e.g., [`std::panic`](https://doc.rust-lang.org/std/panic/index.html) instead of [`core::panic`](https://doc.rust-lang.org/core/panic/index.html).

skejserjensen (changes requested):

server/src/storage/compressed_data_manager.rs
```diff
@@ -0,0 +1,204 @@
...
+
+/// Signed integer since compressed data is inserted first and the remaining bytes are checked after.
+/// This means that the remaining bytes can be negative briefly until compressed data is saved to disk.
+const COMPRESSED_RESERVED_MEMORY_IN_BYTES: isize = 5000;
+
+/// Stores data points compressed as models in memory to batch compressed data before saving it to
+/// Apache Parquet files.
+pub struct CompressedDataManager {
+    /// Path to the folder containing all compressed data managed by the [`StorageEngine`].
+    data_folder_path: PathBuf,
```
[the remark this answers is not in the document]
CGodiksen: It is my understanding that the difference between `PathBuf` and `Path` is the same as `String` and `&str`. Since we want the struct to own the field, we use `PathBuf`. Since `PathBuf` is used extensively, it might be odd to document the difference only here. The alternative would be to document it in all places where it is used but this also seems problematic.

For now I have not made this change but if you think that it is still a good idea, I can do it. Also note that this won't matter if we switch to `object_store` later.

server/src/storage/compressed_data_manager.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    /// Insert `segment` into the in-memory compressed time series buffer.
+    pub fn insert_compressed_data(&mut self, key: String, segment: RecordBatch) {
+        let _span = info_span!("insert_compressed_segment", key = key.clone()).entered();
+        info!(
+            "Inserting batch with {} rows into compressed time series.",
+            segment.num_rows()
+        );
+
+        // Since the compressed segment is already in memory, insert the segment in to the structure
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/compressed_data_manager.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        // If the reserved memory limit is exceeded, save compressed data to disk.
+        if self.compressed_remaining_memory_in_bytes < 0 {
+            self.save_compressed_data();
+        }
+    }
+
+    /// Save [`CompressedTimeSeries`] to disk until the reserved memory limit is no longer exceeded.
+    fn save_compressed_data(&mut self) {
+        info!("Out of memory to store compressed data. Saving compressed data to disk.");
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/compressed_data_manager.rs
```diff
@@ -0,0 +1,204 @@
...
+            // TODO: Maybe create with estimated capacity to avoid reallocation.
+            compressed_data: HashMap::new(),
+            compressed_queue: VecDeque::new(),
+            compressed_remaining_memory_in_bytes: COMPRESSED_RESERVED_MEMORY_IN_BYTES,
+        }
+    }
+
+    /// Insert `segment` into the in-memory compressed time series buffer.
+    pub fn insert_compressed_data(&mut self, key: String, segment: RecordBatch) {
+        let _span = info_span!("insert_compressed_segment", key = key.clone()).entered();
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: The reason links were not added was because this was a normal comment and not a doc comment. I have changed it to a doc comment and added links that include the `s` in the link.

server/src/storage/segment.rs
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: Fixed.

server/src/storage/segment.rs
CGodiksen: Fixed.

server/src/storage/uncompressed_data_manager.rs
CGodiksen: Replaced `Prioritized` with `FIFO`.

server/src/storage/uncompressed_data_manager.rs
CGodiksen: Fixed.

server/src/storage/compressed_data_manager.rs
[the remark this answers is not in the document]
skejserjensen: I was mostly thinking of `PathBuf` as a mutable `Path` and was not aware that it is unsized. But since `Path` is an unsized type I agree that `PathBuf` should be used until we switch to `object_store`.

skejserjensen (approved):

agneborn98 (approved):
PR looks good and changes have been fixed. No comment

PR #26, opened 2022-08-22
Temporarily fix compatibility for SchemaResult

skejserjensen:
This PR adds a temporary workaround for [issue \#2445 in arrow-rs](https://github.com/apache/arrow-rs/issues/2445) by extending [`FlightService.get_schema()`](https://github.com/ModelarData/ModelarDB-RS/blob/bug/serialize-schema/server/src/remote.rs#L178) to always prepend both the [32-bit continuation indicator and the 32-bit little-endian length](https://arrow.apache.org/docs/format/Columnar.html#encapsulated-message-format) to `SchemaResult.schema` so it can be deserialized by the other Apache Arrow Flight implementations. To preserve compatibility with the Rust client (`mdb`), [`execute_command()`](https://github.com/ModelarData/ModelarDB-RS/blob/bug/serialize-schema/client/src/main.rs#L192) have been extended to strip the first eight bytes of `SchemaResult.schema`. I have manually tested the changes with Go (using the code in [Telegraf-Output-Apache-Arrow-Flight](https://github.com/ModelarData/Telegraf-Output-Apache-Arrow-Flight/tree/file-transfer)), Java, Python, and the Rust client (`mdb`) using multiple different schemas. Finally, while slightly outside the scope of the PR, I have moved `DoExchangeStream` before `DoActionStream` so the order of the types should match the order of the methods that use them in `FlightService`.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -189,7 +189,13 @@ fn execute_command(
             let fd = FlightDescriptor::new_path(vec![table_name.to_string()]);
             let request = Request::new(fd);
             rt.block_on(async {
-                let schema_result = fsc.get_schema(request).await?.into_inner();
+                // Both the 32-bit continuation indicator and the 32-bit
+                // little-endian length will be prepended to SchemaResult.schema
+                // for compatibility with other Apache Arrow Flight
+                // implementations until issue
+                // https://github.com/apache/arrow-rs/issues/2445 is fixed.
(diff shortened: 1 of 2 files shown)


agneborn98 (approved):
This looks a lot better and is definitely less "hacky". I have tested it on my computer as well and ran into no issues, so this solution works fine.

CGodiksen (approved):

PR #33, opened 2022-09-18
Update dependencies

skejserjensen:
This pull request primarily updates Apache Arrow DataFusion to [v12.0.0](https://crates.io/crates/datafusion/12.0.0) and fixes all known breaking changes. New versions of Apache Arrow DataFusion usually have breaking changes, so updating to each new release ensures that the number of changes required does not accumulate. The breaking changes in this PR are mainly the split of [`PrimitiveBuilder::new(capacity: usize)` ](https://docs.rs/arrow/21.0.0/arrow/array/struct.PrimitiveBuilder.html#method.new) into [`PrimitiveBuilder::new()`](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveBuilder.html#method.new) and [`PrimitiveBuilder::with_capacity(capacity: usize)`](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveBuilder.html#method.with_capacity). Unfortunately, [Apache Arrow DataFusion v12.0.0](https://crates.io/crates/datafusion/12.0.0) depends on [Apache Arrow v22.0.0](https://crates.io/crates/arrow/22.0.0) which [does not include](https://github.com/apache/arrow-rs/blob/22.0.0/arrow-flight/src/arrow.flight.protocol.rs#L74) the updated [`SchemaResult`](https://github.com/apache/arrow-rs/blob/master/arrow-flight/src/arrow.flight.protocol.rs#L74) which is requied to remove the workaround added by [PR 26](https://github.com/ModelarData/ModelarDB-RS/pull/26). @CGodiksen please double check that changing the value of [`COMPRESSED_SEGMENT_SIZE`](https://github.com/ModelarData/ModelarDB-RS/blob/18a6a7149512fa9e3a90467735ab7a45bb9fae0a/server/src/storage/mod.rs#L315) actually fixed a test and did not hide a bug.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,6 +20,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57e6e951cfbb2db8de1828d49073a113a29fd7117b1596caa781a258c7e38d72"
+dependencies = [
+ "cfg-if",
+ "const-random",
+ "getrandom",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


CGodiksen (approved):
Everything looks fine. I assume the change to `COMPRESSED_SEGMENT_SIZE` is necessary due to a change in the way record batches are stored, specifically the `ArrayData`. After taking a cursory look at the source code it does not seem like the implementation of `column.get_array_memory_size()` is changed.

agneborn98 (approved):

PR #35, opened 2022-09-21
Compressing directly from storage engine

CGodiksen:
This PR includes a minor change to the storage engine that makes it so we instead of queuing finished segments and making them available for retrieval, now just compress the segment directly. In the future this should be changed back and functionality should be added to a compression component that can dynamically allocate threads for compression. Each of these threads should retrieve finished segments from the storage engine and insert the compressed data back into the storage engine when done.

TODOs have been added to revert the storage engine when the above mentioned ideal functionality has been added. Finally, the PR contains a new field on the uncompressed data manager used to make the manager use the old functionality in the tests. This was only added to fix existing tests and thereby avoid changing a significant number of "broken" tests. 

As a minor thing this PR also uses the new `get_array!` macro wherever possible.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -80,7 +80,7 @@ fn main() -> Result<(), String> {
         })?;
         let runtime = Runtime::new().unwrap();
         let mut session = create_session_context();
-        let storage_engine = StorageEngine::new(data_folder_path.clone());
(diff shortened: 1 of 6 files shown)


skejserjensen (approved):

agneborn98 (approved):
I have manually tested this with the telegraf plugin and it looks to be working fine.

PR #42, opened 2022-11-02
Use optimized settings for ArrowWriter

skejserjensen:
This PR configures [ArrowWriter](https://docs.rs/parquet/latest/parquet/arrow/arrow_writer/struct.ArrowWriter.html) with the global [WriterProperties](https://docs.rs/parquet/latest/parquet/file/properties/struct.WriterProperties.html) that provides the best trade-off between ingestion time, query time, and the amount of storage requistoragered according to an extensive search performed by [Evaluate-ModelarDB-Changes](https://github.com/ModelarData/Utilities/tree/main/Evaluate-ModelarDB-Changes) using 1 GiB of real-life data and simple aggregate queries that aggregate the entire ingested data set. The following table shows the set of nondominated results as computed by this [Python script](https://github.com/matthewjwoodruff/pareto.py). For all of the nondominated results the encoding was set to PLAIN and dictionary compression was disabled. The first column specify the compression method used and the level of statistics to include in the Parquet files.

|              | Ingestion Time (s) | Query Time (s) | Data Folder Size (KiB) |
|--------------|--------------------|----------------|------------------------|
| **UNCOMPRESSED** |                    |                |                        |
| NONE         | 24.228             | 1.024          | 658672                 |
| CHUNK        | 24.495             | 1.019          | 658700                 |
| PAGE         | 24.210             | 1.142          | 658728                 |
|              |                    |                |                        |
| **SNAPPY**   |                    |                |                        |
| NONE         | 25.063             | 1.162          | 521452                 |
| CHUNK        | 25.100             | 1.173          | 521488                 |
| PAGE         | 24.720             | 1.160          | 521508                 |
|              |                    |                |                        |
| **GZIP**     |                    |                |                        |
| NONE         | 46.812             | 1.958          | 465804                 |
| CHUNK        | 46.913             | 1.955          | 465832                 |
| PAGE         | 46.760             | 1.952          | 465844                 |
|              |                    |                |                        |
| **BROTLI**   |                    |                |                        |
| NONE         | 26.984             | 3.276          | 474704                 |
| CHUNK        | 27.718             | 3.281          | 474756                 |
| PAGE         | 26.998             | 3.266          | 474780                 |
|              |                    |                |                        |
| **LZ4**      |                    |                |                        |
| NONE         | 25.289             | 1.266          | 515540                 |
| CHUNK        | 25.474             | 1.238          | 515592                 |
| PAGE         | 25.106             | 1.259          | 515608                 |
|              |                    |                |                        |
| **ZSTD**     |                    |                |                        |
| NONE         | 25.459             | 1.252          | 475552                 |
| CHUNK        | 25.812             | 1.241          | 475600                 |
| PAGE         | 25.669             | 1.235          | 475660                 |

Its seems PLAIN is the only encoding that is supported for all column types as either ingestion or query execution failed for all other encodings, so PLAIN is the only encoding we can set globally. Thus, the encoding is purposely not set to PLAIN to allow the [Parquet crate](https://docs.rs/parquet/latest/parquet/) to use per a per column encoding, although it looks like it is just defaulting to plain currently. ZSTD seems to provide almost as good compression as GZIP but it has only a very minor negative impact on ingestion time and query time. While not part of the nondominated set of results, enabling dictionary compression significantly increases the ingestion time without a significant benefit in terms of storage size, so it is still disabled. Finally enabling statistics seems to have no impact on the measured metrics, so PAGE level statistics are enabled in case they are beneficial for pruning. Rerunning the test with the new settings produced the following results.

| ZSTD         | Ingestion Time (s) | Query Time (s) | Data Folder Size (KiB) |
|--------------|--------------------|----------------|------------------------|
| PAGE         | 25.590             | 1.202          | 475636                 |

While the changes to the code is trivial, please check that the use of the default encoding, ZSTD, no dictionary compression, and PAGE level statistics is sound based on the presented results.

### DIFF
--- a/server/src/storage/mod.rs
+++ b/server/src/storage/mod.rs
@@ -32,14 +32,14 @@ use chrono::DateTime;
 use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use datafusion::parquet::arrow::ArrowWriter;
-use datafusion::parquet::basic::Encoding;
+use datafusion::parquet::basic::Compression;
 use datafusion::parquet::errors::ParquetError;
-use datafusion::parquet::file::properties::WriterProperties;
+use datafusion::parquet::file::properties::{EnabledStatistics, WriterProperties};
 use object_store::path::Path as ObjectStorePath;
 use object_store::ObjectMeta;
 
-use crate::metadata::model_table_metadata::ModelTableMetadata;
 use crate::errors::ModelarDbError;
+use crate::metadata::model_table_metadata::ModelTableMetadata;
 use crate::metadata::MetadataManager;
 use crate::storage::compressed_data_manager::CompressedDataManager;
 use crate::storage::segment::FinishedSegment;
@@ -196,7 +196,6 @@ impl StorageEngine {
         Ok(compressed_files)
     }
 
-    // TODO: Test using more efficient encoding. Plain encoding makes it easier to read the files externally.
     /// Write `batch` to an Apache Parquet file at the location given by `file_path`. `file_path`
     /// must use the extension '.parquet'. Return [`Ok`] if the file was written successfully,
     /// otherwise [`ParquetError`].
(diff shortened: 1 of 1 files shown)


CGodiksen (approved):

chrthomsen (approved):
Looks good to me

PR #43, opened 2022-11-07
Removed tokio runtime from context

CGodiksen:
To simplify the process of passing the tokio runtime to sub-components, the runtime has been removed from the context. This means that instead of retrieving the runtime from the context, components that need to carry out async tasks need to receive a reference to the runtime that is created in main. 

Currently there is no large benefit to this change but the need for it becomes more clear when components that rely on the async object_store crate are added (fx. data transfer).

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -109,7 +109,7 @@ fn main() -> Result<(), String> {
         setup_ctrl_c_handler(&context, &runtime);
 
         // Start Interface.
-        remote::start_arrow_flight_server(context, &runtime,9999).map_err(|error| error.to_string())?
+        remote::start_arrow_flight_server(context, &runtime, 9999).map_err(|error| error.to_string())?
     } else {
         // The errors are consciously ignored as the program is terminating.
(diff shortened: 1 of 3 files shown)


server/src/main.rs
```diff
@@ -75,11 +73,13 @@ fn main() -> Result<(), String> {
         fs::create_dir_all(data_folder_path.as_path())
             .map_err(|error| format!("Unable to create {}: {}", &data_folder, error))?;
 
+        // Create a tokio runtime for executing asynchronous tasks.
```
skejserjensen, line 76: Maybe replace "tokio" with "Tokio" for consistency and add an extra sentence to the comment stating why the runtime is not part of the context so we do not get the bright idea in the future to try and make it part of the context again.

skejserjensen (changes requested):

server/src/main.rs
CGodiksen: Fixed.

skejserjensen (approved):

chrthomsen (approved):

PR #45, opened 2022-11-08
Update dependencies

skejserjensen:
This PR primarily updates Apache Arrow DataFusion to [v14.0.0](https://crates.io/crates/datafusion/14.0.0) and fixes all known breaking changes. New versions of Apache Arrow DataFusion usually have breaking changes, so updating to each new release ensures that the number of changes required does not accumulate. The files have purposely only been changed to the degree necessary to fix the breaking changes. The breaking changes this time are mainly caused by ongoing process of splitting Apache Arrow DataFusion into sub-crates, additional methods being added to the [`PhysicalExpr`](https://docs.rs/datafusion-physical-expr/14.0.0/datafusion_physical_expr/trait.PhysicalExpr.html) trait, and [`combine_filters()`](https://docs.rs/datafusion/13.0.0/datafusion/logical_plan/fn.combine_filters.html) seemingly being replaced with [`conjunction`](https://docs.rs/datafusion-optimizer/14.0.0/datafusion_optimizer/utils/fn.conjunction.html).

### DIFF
--- a/server/src/tables.rs
+++ b/server/src/tables.rs
@@ -33,13 +33,13 @@ use datafusion::arrow::array::{
 use datafusion::arrow::datatypes::{ArrowPrimitiveType, Field, Schema, SchemaRef, UInt16Type};
 use datafusion::arrow::error::Result as ArrowResult;
 use datafusion::arrow::record_batch::RecordBatch;
+use datafusion::config::ConfigOptions;
 use datafusion::datasource::object_store::ObjectStoreUrl;
 use datafusion::datasource::{
     datasource::TableProviderFilterPushDown, listing::PartitionedFile, TableProvider, TableType,
 };
 use datafusion::error::{DataFusionError, Result};
 use datafusion::execution::context::{ExecutionProps, SessionState, TaskContext};
-use datafusion::logical_plan::{col, combine_filters, Expr, Operator, ToDFSchema};
 use datafusion::physical_plan::{
     expressions::PhysicalSortExpr, file_format::FileScanConfig, file_format::ParquetExec,
     filter::FilterExec, metrics::BaselineMetrics, metrics::ExecutionPlanMetricsSet,
@@ -47,8 +47,12 @@ use datafusion::physical_plan::{
     SendableRecordBatchStream, Statistics,
 };
 use datafusion::scalar::ScalarValue;
+use datafusion_common::ToDFSchema;
+use datafusion_expr::{col, BinaryExpr, Expr, Operator};
+use datafusion_optimizer::utils;
 use datafusion_physical_expr::planner;
 use futures::stream::{Stream, StreamExt};
+use parking_lot::RwLock;
 
 use crate::metadata::model_table_metadata::ModelTableMetadata;
 use crate::models;
@@ -73,6 +77,8 @@ pub struct ModelTable {
     schema: Arc<Schema>,
     /// Field column to use for queries that do not include fields.
     fallback_field_column: u64,
+    /// Configuration options to use for reading Apache Parquet files.
+    config_options: Arc<RwLock<ConfigOptions>>,
 }
 
 impl ModelTable {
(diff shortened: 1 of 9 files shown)


server/src/tables.rs
```diff
@@ -33,22 +33,26 @@ use datafusion::arrow::array::{
...
     metrics::MetricsSet, DisplayFormatType, ExecutionPlan, Partitioning, RecordBatchStream,
     SendableRecordBatchStream, Statistics,
 };
 use datafusion::scalar::ScalarValue;
+use datafusion_common::ToDFSchema;
+use datafusion_expr::{col, BinaryExpr, Expr, Operator};
+use datafusion_optimizer::utils;
 use datafusion_physical_expr::planner;
 use futures::stream::{Stream, StreamExt};
+use parking_lot::RwLock;
```
CGodiksen, line 55: Do we want to use the [parking_lot](https://crates.io/crates/parking_lot) crate everywhere instead of std::sync?

CGodiksen (changes requested):

server/src/tables.rs
skejserjensen: [parking_lot::RwLock](https://docs.rs/parking_lot/0.12.1/parking_lot/type.RwLock.html) was used as a necessity as [FileScanConfig](https://docs.rs/datafusion/latest/datafusion/physical_plan/file_format/struct.FileScanConfig.html) requires a new argument named [config_options](https://docs.rs/datafusion/latest/datafusion/physical_plan/file_format/struct.FileScanConfig.html#structfield.config_options) with the type `std::sync:::Arc<parkin_lot::RwLock<ConfigOptions>>` in Apache Arrow DataFusion 14, so I had to use [parking_lot::RwLock](https://docs.rs/parking_lot/0.12.1/parking_lot/type.RwLock.html) instead of [std::sync::RwLock ](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html).

However, [parking_lot::RwLock](https://docs.rs/parking_lot/0.12.1/parking_lot/type.RwLock.html) provides a nice summarized comparison to [std::sync::RwLock ](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html) and from my perspective are many of the listed changes attractive improvements such as _Supports atomically downgrading a write lock into a read lock_, _No poisoning, the lock is released normally on panic_, and while it use could be dangerous _Allows raw locking & unlocking without a guard_. So since we are forced to include [parking_lot](https://crates.io/crates/parking_lot) as a dependency, it is fine for me if we just consistently use its synchronization primitives unless they have a disadvantage compared to what is available in [std::sync](https://doc.rust-lang.org/stable/std/sync/index.html).

Should I change the remain [std::sync::RwLock ](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html) to [parking_lot::RwLock](https://docs.rs/parking_lot/0.12.1/parking_lot/type.RwLock.html)  in this PR for consistency? If we decide we want to replace additional synchronization primitives with something from [parking_lot](https://crates.io/crates/parking_lot) I suggest we do that in another PR.

CGodiksen: I think it is fine to make the change in this PR.

skejserjensen: I have replaced [std::sync::RwLock ](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html) with [parking_lot::RwLock](https://docs.rs/parking_lot/0.12.1/parking_lot/type.RwLock.html) as [parking_lot](https://crates.io/crates/parking_lot) does not have an alternative to [std::sync::Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html) as far as I can tell.

server/src/tables.rs
```diff
@@ -232,8 +232,7 @@ impl TableProvider for ModelTable {
         // Request the matching files from the storage engine.
         let mut key_object_metas = {
             // TODO: make the storage engine support multiple parallel readers.
-            // unwrap() is safe as read() only fails if the RwLock is poisoned.
-            let mut storage_engines = self.context.storage_engine.write().unwrap();
+            let mut storage_engines = self.context.storage_engine.write();
```
CGodiksen, line 235: Very minor but maybe change this to `let mut storage_engine`

CGodiksen (changes requested):

server/src/tables.rs
skejserjensen: Fixed.

CGodiksen (approved):

chrthomsen (approved):

PR #50, opened 2022-11-21
Made data transfer async and added action to flush the edge

CGodiksen:
This PR removed all use of the Tokio runtime within the data transfer component (previously used for `runtime.block_on()`) and converts the data transfer component to instead use async/await. This means that the data transfer component can now be invoked from another async function, such as from the Arrow flight endpoints. 

Further, the PR also adds a new action to make it possible to flush the entire edge. This action first flushes uncompressed memory to disk and then flushes all compressed data saved on disk to the remote store if possible.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -127,6 +127,7 @@ fn main() -> Result<(), String> {
         .map_err(|error| format!("Unable to create a MetadataManager: {}", error))?;
     let session = create_session_context(data_folders.query_data_folder);
     let storage_engine = RwLock::new(StorageEngine::new(
+        data_transfer,
         data_folders.local_data_folder,
         metadata_manager.clone(),
         true,
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -41,6 +41,7 @@ use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
 use crate::metadata::MetadataManager;
 use crate::optimizer::model_simple_aggregates;
+use crate::storage::data_transfer::DataTransfer;
 use crate::storage::StorageEngine;
 
 #[global_allocator]
@@ -99,25 +100,14 @@ fn main() -> Result<(), String> {
         Runtime::new().map_err(|error| format!("Unable to create a Tokio Runtime: {}", error))?,
     );
 
-    // Check if a remote data folder was provided and can be accessed. If so, the data transfer
-    // component is initialized. These checks are performed after parse_command_line_arguments()
-    // as the Tokio Runtime is required.
+    // Check if a remote data folder was provided and can be accessed. These checks are performed
+    // after parse_command_line_arguments() as the Tokio Runtime is required.
     let data_transfer = if let Some(remote_data_folder) = data_folders.remote_data_folder {
-        Some(runtime.block_on(async {
-            remote_data_folder
-                .get(&Path::from(""))
-                .await
-                .map_err(|error| error.to_string())?;
-
-            // TODO: Make the transfer batch size in bytes part of the user-configurable settings.
-            storage::data_transfer::DataTransfer::try_new(
-                data_folders.local_data_folder.clone(),
-                remote_data_folder,
-                64 * 1024 * 1024, // 64 MiB.
-            )
-            .await
-            .map_err(|error| error.to_string())
-        })?)
+        Some(check_and_initialize_data_transfer(
+            &runtime,
+            remote_data_folder,
+            data_folders.local_data_folder.clone(),
+        )?)
     } else {
         None
     };
@@ -230,6 +220,29 @@ fn argument_to_remote_object_store(argument: &str) -> Result<Arc<dyn ObjectStore
     } else {
         Err("Remote data folder must be s3://bucket-name.".to_owned())
     }
+}
+
+/// Check if the remote data folder can be accessed. If so, the data transfer component is initialized.
+fn check_and_initialize_data_transfer(
+    runtime: &Runtime,
+    remote_data_folder: Arc<dyn ObjectStore>,
+    local_data_folder: PathBuf,
+) -> Result<DataTransfer, String> {
+    runtime.block_on(async {
+        remote_data_folder
+            .get(&Path::from(""))
+            .await
+            .map_err(|error| error.to_string())?;
+
+        // TODO: Make the transfer batch size in bytes part of the user-configurable settings.
+        DataTransfer::try_new(
+            local_data_folder,
+            remote_data_folder,
+            64 * 1024 * 1024, // 64 MiB.
+        )
+        .await
+        .map_err(|error| error.to_string())
+    })
 }
 
 /// Create a new [`SessionContext`] for interacting with Apache Arrow
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -99,16 +99,28 @@ fn main() -> Result<(), String> {
         Runtime::new().map_err(|error| format!("Unable to create a Tokio Runtime: {}", error))?,
     );
 
-    // Ensure the remote data folder can be accessed. The check is performed after
-    // parse_command_line_arguments() as the Tokio Runtime is required.
-    if let Some(remote_data_folder) = data_folders.remote_data_folder {
-        runtime.block_on(async {
+    // Ensure the remote data folder can be accessed and set up the data transfer component if the
+    // remote data folder exists. The check is performed after parse_command_line_arguments() as the
+    // Tokio Runtime is required.
+    let data_transfer = if let Some(remote_data_folder) = data_folders.remote_data_folder {
+        Some(runtime.block_on(async {
             remote_data_folder
                 .get(&Path::from(""))
                 .await
-                .map_err(|error| error.to_string())
-        })?;
-    }
+                .map_err(|error| error.to_string())?;
+
+            // TODO: Make the transfer batch size in bytes part of the user-configurable settings.
+            storage::data_transfer::DataTransfer::try_new(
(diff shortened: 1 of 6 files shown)


server/src/storage/mod.rs
```diff
@@ -21,7 +21,7 @@ mod compressed_data_manager;
 mod segment;
 mod time_series;
 mod uncompressed_data_manager;
-mod data_transfer;
+pub(crate) mod data_transfer;
```
skejserjensen, line 24: Could this be `pub(super)` if a method is added to `StorageEngine`?

server/src/storage/mod.rs
```diff
@@ -74,11 +75,12 @@ pub struct StorageEngine {
     /// Manager that contains and controls all uncompressed data.
     uncompressed_data_manager: UncompressedDataManager,
     /// Manager that contains and controls all compressed data.
-    compressed_data_manager: CompressedDataManager,
+    pub compressed_data_manager: CompressedDataManager,
```
skejserjensen, line 78: Could this be private if a method is added to `StorageEngine`?

server/src/storage/compressed_data_manager.rs
```diff
@@ -28,13 +28,16 @@ use tracing::info;
...
 use crate::storage::time_series::CompressedTimeSeries;
 use crate::types::{CompressedSchema, Timestamp};
 use crate::StorageEngine;
 
 /// Stores data points compressed as models in memory to batch compressed data before saving it to
 /// Apache Parquet files.
-pub(super) struct CompressedDataManager {
+pub struct CompressedDataManager {
+    /// Component to manage the quantity of saved compressed data and transfer the data when necessary.
+    pub data_transfer: Option<DataTransfer>,
```
skejserjensen, line 40: Could this be `pub(super)` if a method is added to `StorageEngine`?

server/src/storage/compressed_data_manager.rs
```diff
@@ -28,13 +28,16 @@ use tracing::info;
...
 use crate::errors::ModelarDbError;
+use crate::storage::data_transfer::DataTransfer;
 use crate::storage::time_series::CompressedTimeSeries;
 use crate::types::{CompressedSchema, Timestamp};
 use crate::StorageEngine;
 
 /// Stores data points compressed as models in memory to batch compressed data before saving it to
 /// Apache Parquet files.
-pub(super) struct CompressedDataManager {
+pub struct CompressedDataManager {
```
skejserjensen, line 38: Could this be `pub(super)` if a method is added to `StorageEngine`?

server/src/storage/data_transfer.rs
```diff
@@ -109,13 +106,31 @@ impl DataTransfer {
...
-            self.transfer_data(key)?;
+            self.transfer_data(key).await?;
+        }
+
+        Ok(())
+    }
+
+    /// Transfer all compressed files currently in the data folder to the remote object store.
+    /// Return [`Ok`] if all files were transferred successfully, otherwise [`ParquetError`].
+    pub(crate) async fn flush_compressed_files(&mut self) -> Result<(), ParquetError> {
```
skejserjensen, line 127: Maybe add a comment regarding what to do if this method fails, e.g., is it OK to simply call the method again or must half-transferred files be removed from the object store first? Likewise for `transfer_data()`.

skejserjensen (changes requested):

server/src/main.rs
```diff
@@ -99,22 +99,35 @@ fn main() -> Result<(), String> {
...
-                .map_err(|error| error.to_string())
-        })?;
-    }
+                .map_err(|error| error.to_string())?;
+
+            // TODO: Make the transfer batch size in bytes part of the user-configurable settings.
+            storage::data_transfer::DataTransfer::try_new(
+                data_folders.local_data_folder.clone(),
+                remote_data_folder,
+                5000,
```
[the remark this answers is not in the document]
CGodiksen: 5000 was only used for testing. I have changed it to 64 MB for now until further research can be done into optimal batch sizes.

server/src/main.rs
```diff
@@ -99,22 +99,35 @@ fn main() -> Result<(), String> {
         Runtime::new().map_err(|error| format!("Unable to create a Tokio Runtime: {}", error))?,
     );
 
-    // Ensure the remote data folder can be accessed. The check is performed after
-    // parse_command_line_arguments() as the Tokio Runtime is required.
-    if let Some(remote_data_folder) = data_folders.remote_data_folder {
-        runtime.block_on(async {
+    // Ensure the remote data folder can be accessed and set up the data transfer component if the
+    // remote data folder exists. The check is performed after parse_command_line_arguments() as the
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/remote.rs
```diff
@@ -508,6 +509,23 @@ impl FlightService for FlightServiceHandler {
 
             // Confirm the table was created.
             Ok(Response::new(Box::pin(stream::empty())))
+        } else if action.r#type == "FlushEdge" {
```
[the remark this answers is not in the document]
CGodiksen: I personally think that `FlushEdge` makes it clear that the entire edge is flushed of all data.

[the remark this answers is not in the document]
CGodiksen: Alternatively, since the [integration testing](https://github.com/ModelarData/ModelarDB-RS/pull/48) branch introduces an action endpoint called `FlushMemory` that does the same as the first part of this action, we could also change this action to something like `TransferAll` and have it only transfer compressed data.

Then a user of the API could simply call these two endpoints one after the other to achieve the same functionality.

server/src/remote.rs
```diff
@@ -524,7 +542,18 @@ impl FlightService for FlightServiceHandler {
                 .to_owned(),
         };
 
-        let output = stream::iter(vec![Ok(create_command_statement_update_action)]);
+        let flush_edge_action = ActionType {
+            r#type: "FlushEdge".to_owned(),
+            description: "Flush uncompressed data to disk by compressing and saving the data and \
+            flush all compressed data to the remote object store."
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/compressed_data_manager.rs
```diff
@@ -28,13 +28,16 @@ use tracing::info;
...
+    /// Component to manage the quantity of saved compressed data and transfer the data when necessary.
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/remote.rs
```diff
@@ -508,6 +509,23 @@ impl FlightService for FlightServiceHandler {
 
             // Confirm the table was created.
             Ok(Response::new(Box::pin(stream::empty())))
+        } else if action.r#type == "FlushEdge" {
+            let mut storage_engine = self.context.storage_engine.write().await;
+            storage_engine.flush();
+
+            if let Some(data_transfer) = storage_engine
```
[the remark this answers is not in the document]
CGodiksen: Good idea, fixed.

server/src/storage/mod.rs
CGodiksen: Fixed.

server/src/storage/mod.rs
CGodiksen: Fixed.

server/src/storage/compressed_data_manager.rs
CGodiksen: Fixed.

server/src/storage/compressed_data_manager.rs
CGodiksen: Fixed.

server/src/storage/data_transfer.rs
CGodiksen: I have added some minor text that explains that the flush function can be called again if it fails since the data is transferred for each key one at a time. However, I do not think we can add much text about the transfer data function failing since handling partial uploads should be down to the cloud provider. 

Text about how we handle deleting the files on the edge after transferring and how this works when the method fails might be added later when the functionality for that is added.

server/src/main.rs
```diff
@@ -99,22 +99,35 @@ fn main() -> Result<(), String> {
...
+    // Tokio Runtime is required.
+    let data_transfer = if let Some(remote_data_folder) = data_folders.remote_data_folder {
```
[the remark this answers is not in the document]
CGodiksen: Added.

server/src/remote.rs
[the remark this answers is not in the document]
skejserjensen: Personally I would be surprised if an action named `TransferAll` does not transfer the compressed data currently in memory. Let us just stick to `FlushMemory` and `FlushEdge` with their current implementation for now to keep the naming consistent.

server/src/storage/data_transfer.rs
skejserjensen: From the documentation of [`object_store::ObjectStore`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html) it looks like only [`put_multipart()`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#tymethod.put_multipart) needs to be manually cleaned up using [`abort_multipart()`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#tymethod.abort_multipart) as a cleanup method does not seem to exist for [`put()`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#tymethod.put). So the additional documentation you have added is probably fine for now.

skejserjensen (approved):

PR #53, opened 2022-12-15
Refactor the client

skejserjensen:
This PR refactors the client to match the conventions used in the server and adds doc comments.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -66,25 +66,20 @@ fn main() -> Result<(), String> {
         let binary_path = env::current_exe().unwrap();
         let binary_name = binary_path.file_name().unwrap();
         Err(format!(
-            "usage: {} [server_address] [query_file]",
+            "Usage: {} [server_address] [query_file].",
             binary_name.to_str().unwrap()
         ))?;
     }
     let (maybe_host, maybe_query_file_path) = parse_command_line_arguments(args);
 
     // Create the Tokio runtime.
-    let runtime = Runtime::new()
-        .map_err(|error| format!("error: unable to initialize run-time due to {}", error))?;
+    let runtime =
+        Runtime::new().map_err(|error| format!("Unable to create a Tokio Runtime: {}", error))?;
 
     // Connect to the server.
     let host = maybe_host.unwrap_or_else(|| DEFAULT_HOST.to_owned());
-    let flight_service_client = connect(&runtime, &host, DEFAULT_PORT).map_err(|error| {
-        format!(
-            "error: cannot connect to {} due to a {}",
-            host,
-            error.to_string()
-        )
-    })?;
+    let flight_service_client = connect(&runtime, &host, DEFAULT_PORT)
+        .map_err(|error| format!("Cannot connect to {}: {}", host, error))?;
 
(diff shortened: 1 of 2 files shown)


client/src/main.rs
```diff
@@ -12,205 +12,271 @@
...
-                    Err(message) => eprintln!("{}", message),
-                };
-            }
-            Err(message) => eprintln!("error: cannot connect to {} due to a {}", address, message),
-        }
+    let (maybe_host, maybe_query_file_path) = parse_command_line_arguments(args);
+
+    // Create the Tokio runtime.
+    let runtime = Runtime::new()
+        .map_err(|error| format!("error: unable to initialize run-time due to {}", error))?;
```
CGodiksen, line 77: It does not seem like we use the `error:` prefix anywhere in the server. Maybe remove this.

client/src/main.rs
```diff
@@ -12,205 +12,271 @@
...
+
+    // Connect to the server.
+    let host = maybe_host.unwrap_or_else(|| DEFAULT_HOST.to_owned());
+    let flight_service_client = connect(&runtime, &host, DEFAULT_PORT).map_err(|error| {
+        format!(
+            "error: cannot connect to {} due to a {}",
```
CGodiksen, line 83: It does not seem like we use the `error:` prefix anywhere in the server. Maybe remove this.

client/src/main.rs
```diff
@@ -12,205 +12,271 @@
...
+    })?;
+
+    // Execute the queries.
+    if let Some(query_file) = maybe_query_file_path {
+        file(runtime, flight_service_client, &query_file)
     } else {
-        eprintln!("error: unable to initialize run-time");
+        repl(runtime, flight_service_client)
     }
+    .map_err(|error| format!("error: cannot execute queries due to {}", error))
```
CGodiksen, line 95: It does not seem like we use the `error:` prefix anywhere in the server. Maybe remove this.

client/src/main.rs
```diff
@@ -12,205 +12,271 @@
...
     if let Some(mut home) = dirs::home_dir() {
         home.push(history_file_name);
         let _ = editor.append_history(&home);
     }
+
     Ok(())
 }
 
+/// Execute an action, a command, or a query. Returns [`Error`] if the action, command, or query
+/// could not be executed and their result retrieved.
```
CGodiksen, line 207: Should this be `...or their result could not be retrieved.`?

CGodiksen (changes requested):

client/src/main.rs
skejserjensen: Fixed.

client/src/main.rs
skejserjensen: Fixed.

client/src/main.rs
skejserjensen: Fixed.

client/src/main.rs
skejserjensen: Fixed.

CGodiksen (approved):

chrthomsen (approved):

PR #59, opened 2023-01-02
Data transfer integration

CGodiksen:
This PR makes a small change to the compressed data manager in order to add the compressed files to the data transfer component when they are saved to disk. This means that the full workflow from data being ingested to it being saved in the remote object store now works.


server/src/storage/compressed_data_manager.rs
```diff
@@ -296,6 +299,14 @@ impl CompressedDataManager {
             compressed_data_buffer.size_in_bytes, self.compressed_remaining_memory_in_bytes
         );
 
+        // Pass the saved compressed file to the data transfer component if possible.
```
skejserjensen, line 302: Maybe replace "if possible" with a description of what is required for the file to be passed to the data transfer component. For example something like "... if a remote data folder was provided and the size of the file can be retrieved by the data transfer component. If the total size of the files that have been added to the data transfer component have reached the data transfer component's batch size, they are all transferred to the remote data folder."

skejserjensen (changes requested):

server/src/storage/compressed_data_manager.rs
CGodiksen: Fixed.

skejserjensen (approved):

chrthomsen (approved):

PR #62, opened 2023-01-08
Fix pruning of files that contains NaN values

skejserjensen:
This PR updates [`is_compressed_file_within_time_and_value_range()`](https://github.com/ModelarData/ModelarDB-RS/blob/bug/prune-value-range-with-nan/server/src/storage/compressed_data_manager.rs#L348) so it correctly includes/prunes files that store one or more NaN values. The additional checks and tests have been added based on the assumption that a file's value range can only have NaN as the minimum value if the file only contains NaN values as [`arrow::compute::min()`](https://docs.rs/arrow/latest/arrow/compute/fn.min.html) and [`arrow::compute::max()`](https://docs.rs/arrow/latest/arrow/compute/fn.max.html) is used by [`create_time_and_value_range_file_name()`](https://github.com/ModelarData/ModelarDB-RS/blob/bug/prune-value-range-with-nan/server/src/storage/mod.rs#L333) to compute the value range in the filenames and their documentation states that: _For floating point arrays any NaN values are considered to be greater than any other non-null value_.


server/src/storage/mod.rs
```diff
@@ -341,6 +341,9 @@ pub(self) fn create_time_and_value_range_file_name(batch: &RecordBatch) -> Strin
     let max_values = array!(batch, 6, ValueArray);
 
     // unwrap() is safe as None is only returned if all of the values are None.
+    // Both aggregate::min() and aggregate::max() consider NaN to be greater than other non-null
+    // value. So since min_values and max_values cannot contain null, min_value will be NaN if all
```
CGodiksen, line 345: Maybe "than other non-null values." instead.

CGodiksen (changes requested):

CGodiksen (approved):

chrthomsen (approved):

PR #63, opened 2023-01-16
Collect internal statistics

CGodiksen:
This PR includes changes to the storage engine and API to support collecting internal statistics and retrieving them through the Arrow Flight API. Data on the used compressed and uncompressed memory is recorded every time the remaining reserved memory is changed. When uncompressed data is spilled to disk or when compressed files are saved to disk, the used disk space is recorded. Finally, every time a batch of data points is ingested, the number of data points are recorded.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -19,6 +19,7 @@ snmalloc-rs = "0.3.3"
 sqlparser = "0.27.0"
 tokio = { version = "1.22.0", features = ["rt-multi-thread", "signal"] }
 tonic = "0.8.3"
+ringbuf = "0.3.2"
 
 # Log is a dependency so the compile time filters for log and tracing can be set to the same values.
 log = { version = "0.4.17", features = ["max_level_debug", "release_max_level_info"] }
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -14,12 +14,12 @@ futures = "0.3.25"
 object_store = { version = "0.5.1", features = ["aws"] }
 parking_lot = "0.12.1"
 rand = "0.8.5"
+ringbuf = "0.3.2"
 rusqlite = { version = "0.28.0", features = ["bundled"] }
 snmalloc-rs = "0.3.3"
 sqlparser = "0.27.0"
 tokio = { version = "1.22.0", features = ["rt-multi-thread", "signal"] }
 tonic = "0.8.3"
-ringbuf = "0.3.2"
 
 # Log is a dependency so the compile time filters for log and tracing can be set to the same values.
 log = { version = "0.4.17", features = ["max_level_debug", "release_max_level_info"] }
--- a/server/src/remote.rs
+++ b/server/src/remote.rs
@@ -550,8 +550,6 @@ impl FlightService for FlightServiceHandler {
             // Confirm the data was flushed.
             Ok(Response::new(Box::pin(stream::empty())))
         } else if action.r#type == "CollectLogs" {
-            // TODO: Replace the unwraps with map_err to status.
-
             let mut storage_engine = self.context.storage_engine.write().await;
             let logs = storage_engine.collect_logs().await;
 
@@ -597,11 +595,17 @@ impl FlightService for FlightServiceHandler {
             )
             .unwrap();
 
+            // Write the schema and corresponding record batch to a stream.
             let options = IpcWriteOptions::default();
-            let mut writer = StreamWriter::try_new_with_options(vec![], &schema, options).unwrap();
+            let mut writer = StreamWriter::try_new_with_options(vec![], &schema, options)
+                .map_err(|error| Status::internal(error.to_string()))?;
 
-            writer.write(&batch).unwrap();
-            let batch_bytes = writer.into_inner().unwrap();
+            writer
+                .write(&batch)
+                .map_err(|error| Status::internal(error.to_string()))?;
+            let batch_bytes = writer
+                .into_inner()
+                .map_err(|error| Status::internal(error.to_string()))?;
 
             Ok(Response::new(Box::pin(stream::once(async {
                 Ok(arrow_flight::Result { body: batch_bytes })
--- a/server/src/remote.rs
+++ b/server/src/remote.rs
@@ -568,7 +568,22 @@ impl FlightService for FlightServiceHandler {
                 Field::new("values", DataType::List(Box::new(value_field)), false),
             ]);
 
-            Ok(Response::new(Box::pin(stream::once(async { Ok(arrow_flight::Result { body: batch_bytes }) }))))
+            // Extract the data from the logs and insert it into Apache Arrow array builders.
+            let mut log_builder = StringBuilder::new();
+            let mut timestamps_builder = ListBuilder::new(TimestampBuilder::new());
+            let mut values_builder = ListBuilder::new(UInt32Builder::new());
+
+            for (log_name, (timestamps, values)) in logs.iter() {
+                log_builder.append_value(log_name);
+
+                timestamps_builder.values().append_slice(timestamps.values());
+                timestamps_builder.append(true);
+
+                values_builder.values().append_slice(values.values());
+                values_builder.append(true);
+            }
+            
+            Ok(Response::new(Box::pin(stream::empty())))
         } else {
             Err(Status::unimplemented("Action not implemented."))
         }
--- a/server/src/remote.rs
+++ b/server/src/remote.rs
@@ -487,14 +487,14 @@ impl FlightService for FlightServiceHandler {
     /// normal table, and `CREATE MODEL TABLE table_name(...` which creates a model table.
     /// * `FlushMemory`: Flush all data that is currently in memory to disk. This flushes both
     /// uncompressed and compressed data in the storage engine to disk.
-    /// * `FlushEdge`: An extension of the `FlushMemory` action that both flushes all data that is
+    /// * `FlushEdge`: An extension of the `FlushMemory` action that first flushes all data that is
     /// currently in memory to disk and then flushes all compressed data on disk to the remote
     /// object store. Note that data is only transferred to the remote object store if one was
     /// provided when starting the server.
     /// * `CollectMetrics`: Collect internal metrics describing the amount of used memory for
     /// uncompressed and compressed data, used disk space, and ingested data points over time.
     /// Note that the metrics are cleared when collected, meaning that only the metrics
-    /// recorded since the last call to `CollectMetrics` is returned.
+    /// recorded since the last call to `CollectMetrics` are returned.
     async fn do_action(
         &self,
         request: Request<Action>,
--- a/server/src/remote.rs
+++ b/server/src/remote.rs
@@ -33,7 +33,6 @@ use arrow_flight::{
     HandshakeRequest, HandshakeResponse, IpcMessage, PutResult, SchemaAsIpc, SchemaResult, Ticket,
 };
 use datafusion::arrow::array::{ListBuilder, StringBuilder, UInt32Builder};
-use datafusion::arrow::datatypes::{ArrowPrimitiveType, DataType, Field};
 use datafusion::arrow::ipc::writer::StreamWriter;
 use datafusion::arrow::{
     array::ArrayRef, datatypes::Schema, datatypes::SchemaRef, error::ArrowError,
@@ -52,7 +51,7 @@ use crate::metadata::MetadataManager;
 use crate::parser::{self, ValidStatement};
 use crate::storage::{StorageEngine, COMPRESSED_DATA_FOLDER};
 use crate::tables::ModelTable;
-use crate::types::{ArrowTimestamp, TimestampBuilder};
+use crate::types::TimestampBuilder;
 use crate::Context;
 
 /// Start an Apache Arrow Flight server on 0.0.0.0:`port` that pass `context` to
@@ -553,20 +552,6 @@ impl FlightService for FlightServiceHandler {
             let mut storage_engine = self.context.storage_engine.write().await;
             let logs = storage_engine.collect_logs().await;
 
-            // Create the schema for the single record batch that contains all data from the logs.
-            let timestamp_field = Field::new("item", ArrowTimestamp::DATA_TYPE, true);
-            let value_field = Field::new("item", DataType::UInt32, true);
-
-            let schema = Schema::new(vec![
-                Field::new("log", DataType::Utf8, false),
-                Field::new(
-                    "timestamps",
-                    DataType::List(Box::new(timestamp_field)),
-                    false,
-                ),
-                Field::new("values", DataType::List(Box::new(value_field)), false),
-            ]);
-
             // Extract the data from the logs and insert it into Apache Arrow array builders.
             let mut log_builder = StringBuilder::new();
             let mut timestamps_builder = ListBuilder::new(TimestampBuilder::new());
@@ -584,9 +569,11 @@ impl FlightService for FlightServiceHandler {
                 values_builder.append(true);
             }
 
+            let schema = self.context.metadata_manager.metric_schema();
+
             // Finish the builders and create the record batch containing the logs.
             let batch = RecordBatch::try_new(
-                Arc::new(schema.clone()),
+                schema.0.clone(),
                 vec![
                     Arc::new(log_builder.finish()),
                     Arc::new(timestamps_builder.finish()),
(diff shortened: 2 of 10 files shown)


server/src/storage/uncompressed_data_manager.rs
```diff
@@ -59,19 +61,26 @@ pub(super) struct UncompressedDataManager {
     // TODO: This is a temporary field used to fix existing tests. Remove when configuration component is changed.
     /// If this is true, compress finished buffers directly instead of queueing them.
     compress_directly: bool,
+    /// Log of the used uncompressed memory in bytes, updated every time the used memory changes.
+    pub(super) used_uncompressed_memory_log: Log,
+    /// Log of the amount of ingested data points, updated every time a new batch of data is ingested.
+    pub(super) ingested_data_points_log: Log,
+    /// Log of the total used disk space in bytes, updated every time uncompressed data is spilled.
+    pub used_disk_space_log: Arc<RwLock<Log>>,
```
skejserjensen, line 69: As mention above maybe expand the comment for `used_disk_space_log` to state that it is shared for all components in the storage engine and thus must be synchronized.

server/src/storage/mod.rs
```diff
@@ -314,6 +355,48 @@ impl StorageEngine {
...
+        Self {
+            timestamps: TimestampBuilder::new(),
+            values: UInt32Builder::new(),
+            last_value: 0,
+        }
+    }
+
+    /// Add a new entry to the log, where the timestamp is the current milliseconds since the Unix
+    /// epoch and the value is either set directly or based on the last value in the log.
+    fn add_entry(&mut self, value: isize, based_on_last: bool) {
```
skejserjensen, line 379: Maybe rename to `append()` or `append_entry()`. Also maybe replace the `bool` with one method for `based_on_last: false` and another for `based_on_last: true` that both calls a shared method with most of the logic. Although since all three methods would have the same level of visibility, it may be more confusion than the current API.

server/src/storage/uncompressed_data_manager.rs
```diff
@@ -646,23 +698,29 @@ mod tests {
         data_manager.spill_finished_buffer().await;
 
         assert!(remaining_memory < data_manager.uncompressed_remaining_memory_in_bytes);
+        assert_eq!(data_manager.used_uncompressed_memory_log.values.len(), 2);
+
+        // The used disk space log should have an entry for when the uncompressed data manager is
```
skejserjensen, line 703: Maybe add similar comments for other tests were it is not obvious how many and which entries should be in the `Log`s.

skejserjensen (changes requested):

server/src/storage/mod.rs
```diff
@@ -314,6 +355,48 @@ impl StorageEngine {
     }
 }
 
+/// Log used to record changes in specific attributes in the storage engine.
+pub struct Log {
```
skejserjensen, line 359: Maybe include a memory or size limit on `Log`, and if possible use a circular buffer like [RRDtool](https://en.wikipedia.org/wiki/RRDtool), so it will not use an unbounded amount of memory if it is never collected. Alternative, only collect information if it will be requested, e.g., by having a command-line flag or `Action` to enable `Log`. Although this will probably add an excessive amount of complexity to save an insignificant amount of memory.

server/src/remote.rs
```diff
@@ -545,6 +549,67 @@ impl FlightService for FlightServiceHandler {
 
             // Confirm the data was flushed.
             Ok(Response::new(Box::pin(stream::empty())))
+        } else if action.r#type == "CollectLogs" {
+            let mut storage_engine = self.context.storage_engine.write().await;
+            let logs = storage_engine.collect_logs().await;
+
+            // Create the schema for the single record batch that contains all data from the logs.
+            let timestamp_field = Field::new("item", ArrowTimestamp::DATA_TYPE, true);
```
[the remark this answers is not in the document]
CGodiksen: Good idea, fixed.

server/src/storage/mod.rs
```diff
@@ -314,6 +355,48 @@ impl StorageEngine {
...
+    /// Builder consisting of millisecond precision timestamps.
+    timestamps: TimestampBuilder,
+    /// Builder consisting of values.
+    values: UInt32Builder,
+    /// Since the values builder is cleared when the log is finished, the last value is saved separately.
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
CGodiksen: Renamed the method to `append()`. I don't mind the other change but as you mentioned, it might be more confusing so I have kept the current API.

server/src/storage/compressed_data_manager.rs
```diff
@@ -54,6 +55,11 @@ pub(super) struct CompressedDataManager {
     compressed_remaining_memory_in_bytes: isize,
     /// Reference to the schema for compressed data buffers.
     compressed_schema: CompressedSchema,
+    /// Log of the used compressed memory in bytes, updated every time the used memory changes.
+    pub(super) used_compressed_memory_log: Log,
+    /// Log of the total used disk space in bytes, updated every time a new compressed file is
+    /// saved to disk.
+    pub used_disk_space_log: Arc<RwLock<Log>>,
```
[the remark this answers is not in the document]
CGodiksen: I have added a comment to the place where the `used_disk_space_metric` is initialized that explains it to avoid having to explain it in each struct field.

server/src/remote.rs
```diff
@@ -545,6 +549,67 @@ impl FlightService for FlightServiceHandler {
 
             // Confirm the data was flushed.
             Ok(Response::new(Box::pin(stream::empty())))
+        } else if action.r#type == "CollectLogs" {
+            let mut storage_engine = self.context.storage_engine.write().await;
+            let logs = storage_engine.collect_logs().await;
+
+            // Create the schema for the single record batch that contains all data from the logs.
```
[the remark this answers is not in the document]
CGodiksen: Renamed to `metric`.

server/src/remote.rs
```diff
@@ -574,10 +639,20 @@ impl FlightService for FlightServiceHandler {
                 .to_owned(),
         };
 
+        let collect_logs = ActionType {
+            r#type: "CollectLogs".to_owned(),
+            description:
+                "Collect internal logs describing the amount of used memory for uncompressed \
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/uncompressed_data_manager.rs
CGodiksen: See [this](https://github.com/ModelarData/ModelarDB-RS/pull/63#discussion_r1071995640) comment.

server/src/remote.rs
```diff
@@ -545,6 +549,67 @@ impl FlightService for FlightServiceHandler {
...
             Ok(Response::new(Box::pin(stream::empty())))
+        } else if action.r#type == "CollectLogs" {
+            let mut storage_engine = self.context.storage_engine.write().await;
+            let logs = storage_engine.collect_logs().await;
+
+            // Create the schema for the single record batch that contains all data from the logs.
+            let timestamp_field = Field::new("item", ArrowTimestamp::DATA_TYPE, true);
+            let value_field = Field::new("item", DataType::UInt32, true);
+
+            let schema = Schema::new(vec![
```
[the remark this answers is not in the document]
CGodiksen: The current structure of the schema works fine for the way they are used in the UI platform backend. I have left it unchanged for now.

server/src/storage/uncompressed_data_manager.rs
CGodiksen: I think most of the places were clear but I have added a few comments to explain why the used disk space log has one more entry than might be expected.

server/src/storage/mod.rs
CGodiksen: As you mentioned it is quite minor. The intention is that the metrics are collected from all edges using a scheduled task (fx. running every 5 minutes) in the UI platform backend, which should mean that the metrics should never use too much memory. I do however think the used memory is worth considering when further designing how this scheduled task should work so I have added a link to this comment in that work item.

server/src/storage/mod.rs
```diff
@@ -254,6 +262,39 @@ impl StorageEngine {
             .await
     }
 
+    /// Collect and return the logs of used uncompressed/compressed memory, used disk space, and ingested
+    /// data points over time. The logs are returned in tuples with the format (log_name, (timestamps, values)).
+    pub async fn collect_logs(&mut self) -> Vec<(String, (TimestampArray, UInt32Array))> {
+        vec![
+            (
+                "used_uncompressed_memory".to_owned(),
```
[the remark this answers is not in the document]
skejserjensen: Maybe convert `MetricType` to `String` (or more efficiently an int) in `remote.rs` as converting it to `String` in `collect_metrics()` leaves the return type of `collect_metrics()` as `Vec<(String, (TimestampArray, UInt32Array))>` instead of `Vec<(MetricType, (TimestampArray, UInt32Array))>` which would prevent returning an unsupported `MetricType`.

server/src/storage/mod.rs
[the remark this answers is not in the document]
skejserjensen: Maybe "... value instead of simply storing the new value."

server/src/storage/mod.rs
skejserjensen: My main concern is limiting the amount of memory used for collection internal metrics when the system is deployed without the UI of if the UI cannot connect to the system for an extended period of time. In this case the amount of memory used for internal metrics is unbounded and could cause out of memory errors as far as I can tell.

server/src/remote.rs
```diff
@@ -545,6 +548,55 @@ impl FlightService for FlightServiceHandler {
...
+
+                values_builder.values().append_slice(values.values());
+                values_builder.append(true);
+            }
+
+            let schema = self.context.metadata_manager.metric_schema();
+
+            // Finish the builders and create the record batch containing the metrics.
+            let batch = RecordBatch::try_new(
+                schema.0.clone(),
```
skejserjensen, line 576: Unclear why this `.clone()` is needed when `metric_schema()` returns `self.metric_schema.clone()`.

server/src/remote.rs
```diff
@@ -545,6 +548,55 @@ impl FlightService for FlightServiceHandler {
 
             // Confirm the data was flushed.
             Ok(Response::new(Box::pin(stream::empty())))
+        } else if action.r#type == "CollectMetrics" {
```
skejserjensen, line 551: Sorry I did not think about it when performing the initial review, but maybe update the doc comment for `do_action()` as it currently states that `CommandStatementUpdate` is the only supported `Action`. The doc comment also currently contains a lot of detail about `CommandStatementUpdate` so maybe this should be reduced if information about the other `Action`s are added.

skejserjensen (changes requested):

server/src/storage/mod.rs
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/remote.rs
CGodiksen: Removing this `clone()` causes an error since the same schema is used in `StreamWriter::try_new_with_options`. One way to solve this would be to call `self.context.metadata_manager.metric_schema()` twice but I think that would have the exact same effect as calling it once and cloning the result.

server/src/remote.rs
CGodiksen: Good catch, fixed.

server/src/storage/mod.rs
CGodiksen: That is a good point. I agree that this is something that needs to be fixed now and that adding a command-line flag would add an excessive amount of complexity. Following your other suggestion i found a [crate](https://github.com/agerasev/ringbuf) that supports circular buffers and have used it to implement a new version of the `Metric` struct. 

Currently the ring buffer capacity is set so the total memory usage of a single `Metric` should be capped at ~1 MiB. Alternatively we can change this in the future so it scales with the reserved uncompressed/compressed memory.

server/src/remote.rs
```diff
@@ -477,21 +480,21 @@ impl FlightService for FlightServiceHandler {
...
+    /// * `CommandStatementUpdate`: Execute a SQL query containing a command that does not
+    /// return a result. These commands can be `CREATE TABLE table_name(...` which creates a
+    /// normal table, and `CREATE MODEL TABLE table_name(...` which creates a model table.
+    /// * `FlushMemory`: Flush all data that is currently in memory to disk. This flushes both
+    /// uncompressed and compressed data in the storage engine to disk.
+    /// * `FlushEdge`: An extension of the `FlushMemory` action that first flushes all data that is
+    /// currently in memory to disk and then flushes all compressed data on disk to the remote
+    /// object store. Note that data is only transferred to the remote object store if one was
+    /// provided when starting the server.
+    /// * `CollectMetrics`: Collect internal metrics describing the amount of used memory for
```
skejserjensen, line 494: Maybe "...memory used...", "...disk space used...", and "... and the number of data points ingested over time."

server/src/remote.rs
```diff
@@ -477,21 +480,21 @@ impl FlightService for FlightServiceHandler {
...
+    /// uncompressed and compressed data, used disk space, and ingested data points over time.
+    /// Note that the metrics are cleared when collected, meaning that only the metrics
```
skejserjensen, line 496: Maybe replace "meaning that" with "so" or "thus".

server/src/remote.rs
```diff
@@ -477,21 +480,21 @@ impl FlightService for FlightServiceHandler {
...
-    /// * CREATE TABLE company(id INTEGER, name TEXT)
-    /// * CREATE MODEL TABLE wind_turbines(timestamp TIMESTAMP,
-    ///   field_lossless FIELD, field_lossy FIELD(5), tag_one TAG, tag_two TAG)
+    /// Perform a specific action based on the type of the action in `request`. Currently the
+    /// following actions are supported:
+    /// * `CommandStatementUpdate`: Execute a SQL query containing a command that does not
+    /// return a result. These commands can be `CREATE TABLE table_name(...` which creates a
+    /// normal table, and `CREATE MODEL TABLE table_name(...` which creates a model table.
+    /// * `FlushMemory`: Flush all data that is currently in memory to disk. This flushes both
+    /// uncompressed and compressed data in the storage engine to disk.
```
skejserjensen, line 489: I am afraid this could be misunderstood as writing uncompressed data to disk. Maybe we could write something like "This compresses the uncompressed data currently in memory and then flushes all compressed data in the storage engine to disk."

skejserjensen (changes requested):

server/Cargo.toml
```diff
@@ -19,6 +19,7 @@ snmalloc-rs = "0.3.3"
 sqlparser = "0.27.0"
 tokio = { version = "1.22.0", features = ["rt-multi-thread", "signal"] }
 tonic = "0.8.3"
+ringbuf = "0.3.2"
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

server/src/storage/mod.rs
```diff
@@ -410,14 +417,17 @@ impl Metric {
...
-        self.timestamps.append_value(timestamp);
-        self.values.append_value(new_value as u32);
+        self.timestamps.push_overwrite(timestamp);
+        self.values.push_overwrite(new_value as u32);
         self.last_value = new_value;
     }
 
-    /// Finish and reset the internal builders and return the finished Apache Arrow arrays.
+    /// Finish and reset the internal ring buffers and return the timestamps and values as Apache Arrow arrays.
     fn finish(&mut self) -> (TimestampArray, UInt32Array) {
```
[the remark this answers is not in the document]
CGodiksen: Good idea, fixed.

server/src/remote.rs
CGodiksen: Fixed.

server/src/remote.rs
CGodiksen: Fixed.

server/src/remote.rs
CGodiksen: Fixed.

skejserjensen (approved):

chrthomsen (approved):

PR #75, opened 2023-03-02
Added integration tests for list_actions and collect_metrics endpoints

agneborn98:
Additionally, this PR includes a fix for a bug where the generate_values utility would generate a constant time series, and would thereby fail the tests that utilize this helper function.

### DIFF
--- a/crates/modelardb_compression/src/lib.rs
+++ b/crates/modelardb_compression/src/lib.rs
@@ -14,7 +14,7 @@
  */
 
 //! Compress batches of sorted data points represented by a [`TimestampArray`] and a [`ValueArray`]
-//! using the model types in [`models`](crate::models) to produce compressed segments.
+//! using the model types in [`models`](models) to produce compressed segments.
 
 pub mod models;
 
@@ -45,7 +45,7 @@ pub const GORILLA_MAXIMUM_LENGTH: usize = 50;
 /// Compress `uncompressed_timestamps` using a start time, end time, and a
 /// sampling interval if regular and delta-of-deltas followed by a variable
 /// length binary encoding if irregular. `uncompressed_values` is compressed
-/// within `error_bound` using the model types in [`models`](crate::models).
+/// within `error_bound` using the model types in [`models`](models).
 /// Assumes `uncompressed_timestamps` and `uncompressed_values` are sorted
 /// according to `uncompressed_timestamps`. Returns
 /// [`CompressionError`](ModelarDbError::CompressionError) if
@@ -233,7 +233,7 @@ fn can_models_be_merged(
 }
 
 /// A compressed segment being built from an uncompressed segment using the
-/// model types in [`models`](crate::models). Each of the model types is used to
+/// model types in [`models`](models). Each of the model types is used to
 /// fit models to the data points, and then the model that uses the fewest
 /// number of bytes per value is selected.
 struct CompressedSegmentBuilder<'a> {
@@ -952,19 +952,19 @@ pub mod test_util {
     /// will match `timestamps` and their structure will match [`StructureOfValues`]. If `Random` is
     /// selected, `min` and `max` is the range of values which can be generated. If `AlmostLinear`
     /// is selected, `min` and `max` is the maximum and minimum change that should be applied from
-    /// one value to the next. Returns the generated values as a [`Vec`].
+    /// one value to the next. Returns the generated values as a [`Vec<f32>`].
     pub fn generate_values(
         timestamps: &[i64],
         data_type: StructureOfValues,
         min: Option<f32>,
         max: Option<f32>,
     ) -> Vec<f32> {
+        let mut randomizer = thread_rng();
         match data_type {
             // Generates almost linear data.
             StructureOfValues::AlmostLinear => {
                 let a: i64 = thread_rng().gen_range(-10..10);
                 let b: i64 = thread_rng().gen_range(1..50);
-                let mut randomizer = thread_rng();
 
                 timestamps
                     .iter()
@@ -976,7 +976,10 @@ pub mod test_util {
             }
             // Generates linear data.
             StructureOfValues::Linear => {
-                let a: i64 = thread_rng().gen_range(-10..10);
+                let mut a: i64 = 0;
+                while a == 0{
+                    a = thread_rng().gen_range(-10..10);
+                }
                 let b: i64 = thread_rng().gen_range(1..50);
 
                 timestamps
--- a/crates/modelardb_compression/src/lib.rs
+++ b/crates/modelardb_compression/src/lib.rs
@@ -963,7 +963,10 @@ pub mod test_util {
         match data_type {
             // Generates almost linear data.
             StructureOfValues::AlmostLinear => {
-                let a: i64 = thread_rng().gen_range(-10..10);
+                let mut a: i64 = 0;
+                while a == 0{
+                    a = thread_rng().gen_range(-10..10);
+                }
                 let b: i64 = thread_rng().gen_range(1..50);
 
                 timestamps
--- a/crates/modelardb_server/tests/integration_test.rs
+++ b/crates/modelardb_server/tests/integration_test.rs
@@ -623,11 +623,10 @@ fn flush_data_to_disk(runtime: &Runtime, flight_service_client: &mut FlightServi
         r#type: "FlushMemory".to_owned(),
         body: Bytes::new(),
     };
-    let request = Request::new(action);
 
     runtime.block_on(async {
         flight_service_client
-            .do_action(request)
+            .do_action(Request::new(action))
             .await
             .expect("Could not flush data.");
     })
--- a/crates/modelardb_server/tests/integration_test.rs
+++ b/crates/modelardb_server/tests/integration_test.rs
@@ -35,13 +35,13 @@ use datafusion::arrow::datatypes::{DataType, Field, Schema, TimeUnit::Millisecon
 use datafusion::arrow::ipc::convert;
 use datafusion::arrow::ipc::writer::{DictionaryTracker, IpcDataGenerator, IpcWriteOptions};
 use datafusion::arrow::record_batch::RecordBatch;
-use futures::stream;
+use futures::{stream, StreamExt};
 use serial_test::serial;
 use sysinfo::{Pid, PidExt, System, SystemExt};
 
 use tokio::runtime::Runtime;
 use tonic::transport::Channel;
-use tonic::Request;
+use tonic::{Request};
 
 /// The different types of tables used in the integration tests.
 enum TableType {
@@ -164,6 +164,68 @@ fn test_can_get_schema() {
             Field::new("tag", DataType::Utf8, false)
         ])
     );
+
+    stop_modelardbd(flight_server);
+}
+
+#[test]
+#[serial]
+fn test_can_list_actions(){
+    let temp_dir = tempfile::tempdir().expect("Could not create a directory.");
+    let flight_server = start_modelardbd(temp_dir.path());
+
+    let runtime = Runtime::new().expect("Unable to initialize runtime.");
+    let mut flight_service_client = create_apache_arrow_flight_service_client(&runtime, HOST, PORT)
+        .expect("Cannot connect to flight service client.");
+
+    let mut actions = Vec::new();
+    runtime.block_on(async {
+        let mut response = flight_service_client
+            .list_actions(Request::new(arrow_flight::Empty {}))
+            .await
+            .expect("Could not retrieve actions.")
+            .into_inner();
+
+
+        while let Some(action) = response.next().await{
+            actions.push(action.expect("Could not append action to vector.").r#type)
+        }
+    });
+
+    assert_eq!(actions, vec!["CommandStatementUpdate", "FlushMemory", "FlushEdge", "CollectMetrics"]);
+
+    stop_modelardbd(flight_server);
+}
+
+#[test]
+#[serial]
+fn test_can_collect_metrics(){
+    let temp_dir = tempfile::tempdir().expect("Could not create a directory.");
+    let flight_server = start_modelardbd(temp_dir.path());
+
+    let runtime = Runtime::new().expect("Unable to initialize runtime.");
+    let mut flight_service_client = create_apache_arrow_flight_service_client(&runtime, HOST, PORT)
+        .expect("Cannot connect to flight service client.");
+
+    let action = Action {
+        r#type: "CollectMetrics".to_owned(),
+        body: Bytes::new(),
+    };
+
+    let mut metrics = Vec::new();
+    runtime.block_on(async {
+        let mut response = flight_service_client
+            .do_action(Request::new(action))
+            .await
+            .expect("Could not collect metrics.")
+            .into_inner();
+
+        while let Some(metric) = response.next().await{
+            metrics.push(metric.expect("Could not append metric to vector."));
+        }
+    });
+
+    assert!(!metrics.is_empty());
 
     stop_modelardbd(flight_server);
 }


crates/modelardb_compression/src/lib.rs
```diff
@@ -14,7 +14,7 @@
  */
 
 //! Compress batches of sorted data points represented by a [`TimestampArray`] and a [`ValueArray`]
-//! using the model types in [`models`](crate::models) to produce compressed segments.
+//! using the model types in [`models`](models) to produce compressed segments.
```
skejserjensen, line 17: Maybe remove "(models)" as it should not be necessary [for rustdoc to create the correct link](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html).

crates/modelardb_compression/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub const GORILLA_MAXIMUM_LENGTH: usize = 50;
 /// Compress `uncompressed_timestamps` using a start time, end time, and a
 /// sampling interval if regular and delta-of-deltas followed by a variable
 /// length binary encoding if irregular. `uncompressed_values` is compressed
-/// within `error_bound` using the model types in [`models`](crate::models).
+/// within `error_bound` using the model types in [`models`](models).
```
skejserjensen, line 48: Maybe remove "(models)" as it should not be necessary [for rustdoc to create the correct link](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html).

crates/modelardb_compression/src/lib.rs
```diff
@@ -233,7 +233,7 @@ fn can_models_be_merged(
 }
 
 /// A compressed segment being built from an uncompressed segment using the
-/// model types in [`models`](crate::models). Each of the model types is used to
+/// model types in [`models`](models). Each of the model types is used to
```
skejserjensen, line 236: Maybe remove "(models)" as it should not be necessary [for rustdoc to create the correct link](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html).

crates/modelardb_compression/src/lib.rs
```diff
@@ -952,19 +952,22 @@ pub mod test_util {
...
         min: Option<f32>,
         max: Option<f32>,
     ) -> Vec<f32> {
+        let mut randomizer = thread_rng();
         match data_type {
             // Generates almost linear data.
             StructureOfValues::AlmostLinear => {
-                let a: i64 = thread_rng().gen_range(-10..10);
+                let mut a: i64 = 0;
+                while a == 0{
```
skejserjensen, line 967: Missing space between `0` and `{`, try reformatting the file with `rustfmt` and if that does not work add the space manually.

crates/modelardb_compression/src/lib.rs
```diff
@@ -952,19 +952,22 @@ pub mod test_util {
...
         data_type: StructureOfValues,
         min: Option<f32>,
         max: Option<f32>,
     ) -> Vec<f32> {
+        let mut randomizer = thread_rng();
         match data_type {
             // Generates almost linear data.
             StructureOfValues::AlmostLinear => {
-                let a: i64 = thread_rng().gen_range(-10..10);
+                let mut a: i64 = 0;
```
skejserjensen, line 966: Maybe add a comment with the reason why `a == 0` is not allowed.

crates/modelardb_compression/src/lib.rs
```diff
@@ -976,7 +979,10 @@ pub mod test_util {
             }
             // Generates linear data.
             StructureOfValues::Linear => {
-                let a: i64 = thread_rng().gen_range(-10..10);
+                let mut a: i64 = 0;
```
skejserjensen, line 982: Maybe add a comment with the reason why `a == 0` is not allowed.

crates/modelardb_compression/src/lib.rs
```diff
@@ -976,7 +979,10 @@ pub mod test_util {
...
+                while a == 0{
```
skejserjensen, line 983: Missing space between `0` and `{`, try reformatting the file with `rustfmt` and if that does not work add the space manually.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -35,13 +35,13 @@ use datafusion::arrow::datatypes::{DataType, Field, Schema, TimeUnit::Millisecon
...
 use datafusion::arrow::record_batch::RecordBatch;
-use futures::stream;
+use futures::{stream, StreamExt};
 use serial_test::serial;
 use sysinfo::{Pid, PidExt, System, SystemExt};
 
 use tokio::runtime::Runtime;
 use tonic::transport::Channel;
-use tonic::Request;
+use tonic::{Request};
```
skejserjensen, line 44: The added `{ }` does not seem necessary and should be removed unless they are required.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -168,6 +168,68 @@ fn test_can_get_schema() {
...
+        .expect("Cannot connect to flight service client.");
+
+    let mut actions = Vec::new();
+    runtime.block_on(async {
+        let mut response = flight_service_client
+            .list_actions(Request::new(arrow_flight::Empty {}))
+            .await
+            .expect("Could not retrieve actions.")
+            .into_inner();
+
```
skejserjensen, line 188: The extra newline does not seem necessary and should be removed unless it is required.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -168,6 +168,68 @@ fn test_can_get_schema() {
...
+
+        while let Some(action) = response.next().await{
```
skejserjensen, line 190: `stream` implements [`collect()`](https://docs.rs/futures/latest/futures/stream/struct.Collect.html) so it should not be necessary to create a `mut Vec` as it should be possible to simply collect the result into an immutable `Vec` [like in `compressed_data_manager`](https://github.com/ModelarData/ModelarDB-RS/blob/master/crates/modelardb_server/src/storage/compressed_data_manager.rs#L267).

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -168,6 +168,68 @@ fn test_can_get_schema() {
...
+
+    let mut metrics = Vec::new();
+    runtime.block_on(async {
+        let mut response = flight_service_client
+            .do_action(Request::new(action))
+            .await
+            .expect("Could not collect metrics.")
+            .into_inner();
+
+        while let Some(metric) = response.next().await{
```
skejserjensen, line 223: `stream` implements [`collect()`](https://docs.rs/futures/latest/futures/stream/struct.Collect.html) so it should not be necessary to create a `mut Vec` as it should be possible to simply collect the result into an immutable `Vec` [like in `compressed_data_manager`](https://github.com/ModelarData/ModelarDB-RS/blob/master/crates/modelardb_server/src/storage/compressed_data_manager.rs#L267).

skejserjensen (changes requested):

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_server/tests/integration_test.rs
agneborn98: Forgot to run rustfmt. Fixed.

crates/modelardb_server/tests/integration_test.rs
agneborn98: Fixed.

crates/modelardb_server/tests/integration_test.rs
agneborn98: Like this?

crates/modelardb_server/tests/integration_test.rs
agneborn98: Like this?

crates/modelardb_compression/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub const GORILLA_MAXIMUM_LENGTH: usize = 50;
 /// Compress `uncompressed_timestamps` using a start time, end time, and a
 /// sampling interval if regular and delta-of-deltas followed by a variable
 /// length binary encoding if irregular. `uncompressed_values` is compressed
-/// within `error_bound` using the model types in [`models`](models).
+/// within `error_bound` using the model types in [`models`].
```
skejserjensen, line 48: As the lines in the middle of the doc comment has been shortened, the doc comment should be reformatted so each line fits as many words as possible without exceeding the maximum line width. Editors often has functionality to automate this task, e.g., Emacs provides the `fill-paragraph` command while Neovim and Vim provides similar functionality through the `gq` keybinding.

crates/modelardb_compression/src/lib.rs
```diff
@@ -233,7 +233,7 @@ fn can_models_be_merged(
 }
 
 /// A compressed segment being built from an uncompressed segment using the
-/// model types in [`models`](models). Each of the model types is used to
+/// model types in [`models`]. Each of the model types is used to
```
skejserjensen, line 236: As the lines in the middle of the doc comment has been shortened, the doc comment should be reformatted so each line fits as many words as possible without exceeding the maximum line width. Editors often has functionality to automate this task, e.g., Emacs provides the `fill-paragraph` command while Neovim and Vim provides similar functionality through the `gq` keybinding.

skejserjensen (changes requested):

crates/modelardb_server/tests/integration_test.rs
skejserjensen: Yes, but it unclear why the variables and`response` and `metrics` are introduced, maybe replace them `flight_service_client.do_action(...` followed by each operation on the resulting stream on a separate line. By replacing `.await;` with `.await` the result of the transformations will be returned from `block_on()`.

crates/modelardb_server/tests/integration_test.rs
skejserjensen: Yes, but it is unclear why the variables and`response` and `actions` are introduced, maybe replace them with `flight_service_client.list_actions(...` followed by each operation on the resulting stream on a separate line. By replacing `.await;` with `.await` the result of the transformations will be returned from `block_on()`.

crates/modelardb_server/tests/integration_test.rs
agneborn98: Ahh, I see. Fixed.

crates/modelardb_server/tests/integration_test.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
agneborn98: Fixed.

crates/modelardb_compression/src/lib.rs
skejserjensen: This seems to be fixed so I have marked it as resolved.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -168,6 +168,73 @@ fn test_can_get_schema() {
...
+            "FlushMemory",
+        ]
+    );
+
+    stop_modelardbd(flight_server);
+}
+
+#[test]
+#[serial]
+fn test_can_collect_metrics() {
```
CGodiksen, line 211: It seems odd to only test the case where there are no metrics that have been recorded yet. The recording of metrics is itself tested in the storage engine but it would be relevant to test that this is correctly returned when calling the action. It can be a little difficult to force each of the metrics to have entries, specifically `used_compressed_memory` or `used_disk_space` but it should be possible to check that `used_uncompressed_memory` and `ingested_data_points` returns the correct metrics.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -168,6 +168,73 @@ fn test_can_get_schema() {
...
+    let metrics = runtime.block_on(async {
+        flight_service_client
+            .do_action(Request::new(Action {
+                r#type: "CollectMetrics".to_owned(),
+                body: Bytes::new(),
+            }))
+            .await
+            .expect("Could not collect metrics.")
+            .into_inner()
+            .map(|metric| metric.expect("").body)
```
CGodiksen, line 228: Maybe add a message to this call to `expect()`.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -168,6 +168,73 @@ fn test_can_get_schema() {
...
 
+#[test]
+#[serial]
+fn test_can_list_actions() {
+    let temp_dir = tempfile::tempdir().expect("Could not create a directory.");
+    let flight_server = start_modelardbd(temp_dir.path());
+
+    let runtime = Runtime::new().expect("Unable to initialize runtime.");
+    let mut flight_service_client = create_apache_arrow_flight_service_client(&runtime, HOST, PORT)
+        .expect("Cannot connect to flight service client.");
```
CGodiksen, line 179: This is a very minor detail but have we made any decisions on what tense to use for error messages? I noticed that we mix present tense `Cannot` and past tense `Could not` quite often. Do you have an opinion on this @skejserjensen?

CGodiksen (changes requested):

crates/modelardb_server/tests/integration_test.rs
skejserjensen: It is honestly not something I have thought about, but I think past tense makes the most sense for errors and log messages as the user reads them after the error has occurred and not while the error is occurring. However, after looking through some of the code I have written I clearly use both tenses interchangeably. What do you think @chrthomsen?

chrthomsen: Good question... We're not the only ones wondering: https://softwareengineering.stackexchange.com/questions/270899/exception-messages-and-event-logging-which-tense-to-use

I agree that past tense generally is good to use when an attempt was made and it failed. But on the other hand, a message like "File did not exist" also seems weird compared to "File does not exist". So perhaps we should just use the message that most clearly describes the situation?

agneborn98: I think I agree with @chrthomsen, that in most cases it should be past tense but there are exceptions like the file one he mentioned (and probably others).

crates/modelardb_server/tests/integration_test.rs
agneborn98: Added work item in Azure for this.

crates/modelardb_server/tests/integration_test.rs
agneborn98: I have changed the occurrences of "cannot" in the integration_test.rs file, but I have not gone through the entire project.

CGodiksen (approved):

PR #77, opened 2023-03-10
Support error bounds in the form of floats

skejserjensen:
This PR allows error bounds for `FIELD`s to be specified as `f32` instead of `u64`.

### DIFF
--- a/crates/modelardb_server/src/parser.rs
+++ b/crates/modelardb_server/src/parser.rs
@@ -114,11 +114,11 @@ impl ModelarDbDialect {
                     // An error bound may also be specified for field columns.
                     let error_bound = if parser.peek_token() == Token::LParen {
                         parser.expect_token(&Token::LParen)?;
-                        let error_bound = parser.parse_literal_uint()?;
+                        let error_bound = self.parse_positive_literal_f32(parser)?;
                         parser.expect_token(&Token::RParen)?;
                         error_bound
                     } else {
-                        0
+                        0.0
                     };
 
                     // An error bound column option does not exist, so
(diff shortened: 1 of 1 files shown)


CGodiksen (approved):

crates/modelardb_server/src/parser.rs
```diff
@@ -176,6 +176,20 @@ impl ModelarDbDialect {
         }
     }
 
+    /// Return its value as a [`f32`] if the next [`Token`] is a
+    /// [`Token::Number`], otherwise a [`ParserError`] is returned.
+    fn parse_positive_literal_f32(&self, parser: &mut Parser) -> Result<f32, ParserError> {
+        let token_with_location = parser.next_token();
+        match token_with_location.token {
+            Token::Number(maybe_f32, _) => maybe_f32.parse::<f32>().map_err(|error| {
```
chrthomsen, line 184: Is this locale-dependent, i.e., 0.5 for en_US, but 0,5 for da_DK?

crates/modelardb_server/src/parser.rs
```diff
@@ -681,7 +695,7 @@ mod tests {
 
     #[test]
     fn test_tokenize_and_parse_create_model_table() {
-        let sql = "CREATE MODEL TABLE table_name(timestamp TIMESTAMP, field_one FIELD, field_two FIELD(10), tag TAG)";
+        let sql = "CREATE MODEL TABLE table_name(timestamp TIMESTAMP, field_one FIELD, field_two FIELD(10.5), tag TAG)";
```
chrthomsen, line 698: Can't this fail for locale da_DK if float parsing is locale-dependent?

crates/modelardb_server/src/parser.rs
skejserjensen: Based on the documentation it seems only `en_US` is supported as `f32`s can only be parsed if they follow this [grammar](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.from_str).

crates/modelardb_server/src/parser.rs
skejserjensen: Based on the documentation it seems only `en_US` is supported as `f32`s can only be parsed if they follow this [grammar](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.from_str).

skejserjensen: Based on the documentation it seems only `en_US` is supported as `f32`s can only be parsed if they follow this [grammar](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.from_str).

crates/modelardb_server/src/parser.rs
chrthomsen: OK, good.

crates/modelardb_server/src/parser.rs
chrthomsen: Good, this is not a problem then

chrthomsen (approved):

PR #83, opened 2023-03-24
New action for updating remote object store

CGodiksen:
This PR includes a new action to update the remote object store while the server is running. Currently both Amazon S3 and Azure Blob Storage object stores are supported. 

It should be noted that cloud nodes and edge nodes should handle the update slightly differently. Both node types should update the remote data folder (the folder data is transferred to) but for cloud nodes the query data folder should also be updated. This update is problematic since the query data folder is used in the Datafusion session context for cloud nodes. Putting a `RwLock` around the `session` within `Context` did not work and putting a `RwLock` around the entire `Context` is both too heavy-handed and brings about some problems with registering tables.

A work item has been created to solve this issue later. It might be relevant to look at the issue again after more work has been done within moving the system towards `async` in general and when the new cloud architecture implementation has been started. 

Finally the PR also includes a minor change to the compressed file names, specifically has the port of the server been added to the end of the file name to uniquely identify the edge when transferring.  Note that the port is only used for simplicity for now and should be replaced by a different ID when the new cloud architecture is implemented.


crates/modelardb_server/src/remote.rs
```diff
@@ -669,6 +752,37 @@ impl FlightService for FlightServiceHandler {
...
+            storage_engine
+                .update_remote_data_folder(object_store)
+                .await
+                .map_err(|error| {
+                    Status::internal(format!("Could not update remote data folder: {error}"))
+                })?;
+
+            // If on a cloud node, both the remote data folder and the query data folder should be updated.
+            if self.context.metadata_manager.node_type == NodeType::Cloud {
+                // TODO: The query data folder should be updated in the session context.
```
skejserjensen, line 778: Maybe move the detailed TODO messages to the task and keep only a single TODO that states that this feature should be implemented so the detailed information and discussion is in one place.

crates/modelardb_server/src/remote.rs
```diff
@@ -669,6 +752,37 @@ impl FlightService for FlightServiceHandler {
...
+            let mut storage_engine = self.context.storage_engine.write().await;
+            storage_engine
+                .update_remote_data_folder(object_store)
+                .await
+                .map_err(|error| {
+                    Status::internal(format!("Could not update remote data folder: {error}"))
+                })?;
+
+            // If on a cloud node, both the remote data folder and the query data folder should be updated.
+            if self.context.metadata_manager.node_type == NodeType::Cloud {
```
skejserjensen, line 777: Maybe move this to line 756 and immediately return an error if this is a cloud node, otherwise it becomes possible to have a different remote object store for the data transfer component and the query engine.

crates/modelardb_server/src/remote.rs
```diff
@@ -669,6 +752,37 @@ impl FlightService for FlightServiceHandler {
                     body: batch_bytes.into(),
                 })
             }))))
+        } else if action.r#type == "UpdateRemoteObjectStore" {
+            // If the type of the new remote object store is not "s3" or "azureblobstorage", return an error.
+            let (object_store_type, offset_data) = extract_argument(&action.body)?;
+
+            let object_store = match object_store_type {
```
skejserjensen, line 759: Should it be possible to pass something like `None` to disable data transfer? This use case may be too niche to spend development time on it.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -95,6 +96,8 @@ const APACHE_PARQUET_FILE_SIGNATURE: &[u8] = &[80, 65, 82, 49]; // PAR1.
 /// than the requested due to internal alignment when allocating memory for the two array builders.
 const UNCOMPRESSED_DATA_BUFFER_CAPACITY: usize = 64 * 1024;
 
+const TRANSFER_BATCH_SIZE_IN_BYTES: usize = 64 * 1024 * 1024; // 64 MiB;
```
skejserjensen, line 99: A doc comment should be added to this `const` like is already done for the other `const`s.

crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -876,7 +877,7 @@ mod tests {
     /// Return a full path to the file with `file_name`.
     fn format_path(file_name: &str) -> String {
         format!(
-            "{COMPRESSED_DATA_FOLDER}/{TABLE_NAME}/{COLUMN_INDEX}/{file_name}_{}.parquet",
+            "{COMPRESSED_DATA_FOLDER}/{TABLE_NAME}/{COLUMN_INDEX}/{file_name}_{}_{PORT}.parquet",
```
skejserjensen, line 880: The `const`s should be referenced the same way, so either imported completely like `PORT` or by importing the module like `storage::TEST_UUID`.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -344,6 +347,30 @@ impl StorageEngine {
         ]
     }
 
+    /// Update the remote data folder used to transfer data in the data transfer component.
```
skejserjensen, line 350: Seem something is missing from this sentence, maybe `... used to transfer data to in the...`.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -597,9 +625,12 @@ pub(self) fn create_time_and_value_range_file_name(batch: &RecordBatch) -> Strin
         Uuid::new_v4()
     };
 
+    // TODO: Use part of the UUID or the entire Apache Arrow Flight URL to identify the edge.
```
skejserjensen, line 628: The port will be the same for each edge node, so for now maybe use something like the edge node's [`name`](https://docs.rs/sysinfo/latest/sysinfo/trait.SystemExt.html#tymethod.name) or [`host_name`](https://docs.rs/sysinfo/latest/sysinfo/trait.SystemExt.html#tymethod.host_name) from the [`sysinfo`](https://crates.io/crates/sysinfo) crate. The name should probably be cached instead of retrieving it from the system each time.

skejserjensen (changes requested):

crates/modelardb_server/src/main.rs
```diff
@@ -41,6 +41,15 @@ use crate::storage::StorageEngine;
 #[global_allocator]
 static ALLOC: snmalloc_rs::SnMalloc = snmalloc_rs::SnMalloc;
 
+const PORT: i16 = 9999;
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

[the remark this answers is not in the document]
CGodiksen: Fixed.

crates/modelardb_server/src/main.rs
```diff
@@ -41,6 +41,15 @@ use crate::storage::StorageEngine;
...
+
+/// The different possible node types, assigned when the node is started.
+#[derive(Clone, PartialEq)]
```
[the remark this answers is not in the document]
CGodiksen: It seems that `Eq` requires that it also derives `PartialEq`. It also seems that `Eq` is "only informing the compiler that this is an equivalence relation rather than a partial equivalence relation" while `PartialEq` is for supporting actual equality operations. I have added `Eq` to the derive and kept `PartialEq` as well.

crates/modelardb_server/src/main.rs
```diff
@@ -41,6 +41,15 @@ use crate::storage::StorageEngine;
...
+pub enum NodeType {
```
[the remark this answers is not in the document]
CGodiksen: Changed to `ServerMode`.

crates/modelardb_server/src/remote.rs
```diff
@@ -140,6 +144,77 @@ async fn send_flight_data(
...
+
+    Ok((argument, remaining_bytes))
+}
+
+/// Check that the connection is valid with the given arguments. Note that an "object not found"
+/// error is returned even if the connection is valid. We therefore check for the specific error.
+async fn check_object_store(object_store: Arc<dyn ObjectStore>) -> Result<Arc<dyn ObjectStore>, Status> {
+    match object_store.get(&Path::from("")).await {
+        Ok(_) => Ok(object_store),
+        Err(ref error) if error.to_string().contains("Object at location  not found") => {
```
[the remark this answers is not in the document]
CGodiksen: This was due to the path being given to `object_store.get()` being "" but since we check the error type now, it has been removed.

crates/modelardb_server/src/remote.rs
```diff
@@ -140,6 +144,77 @@ async fn send_flight_data(
...
+        .map_err(|error| Status::invalid_argument(error.to_string()))?;
+    let remaining_bytes = &data[(size + 2)..];
+
+    Ok((argument, remaining_bytes))
+}
+
+/// Check that the connection is valid with the given arguments. Note that an "object not found"
+/// error is returned even if the connection is valid. We therefore check for the specific error.
+async fn check_object_store(object_store: Arc<dyn ObjectStore>) -> Result<Arc<dyn ObjectStore>, Status> {
+    match object_store.get(&Path::from("")).await {
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

crates/modelardb_server/src/remote.rs
CGodiksen: Fixed.

crates/modelardb_server/src/remote.rs
CGodiksen: Fixed.

crates/modelardb_server/src/remote.rs
CGodiksen: We discussed making it possible to delete the object store to disable data transfer but it was determined that it was too niche to include in the main version of the system. Instead it should only be added if a customer specifically requests it.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: Fixed.

crates/modelardb_server/src/storage/compressed_data_manager.rs
CGodiksen: Importing them both completely to make the `format!` cleaner.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: Fixed.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: The reason the port is used for now is that it is available in the platform, which is the part that needs to identify the edge during data transfers. The port will be unique for now since the only place we will have multiple edges is in the demo environment where each edge will run on a different port.

Using the name or any unique ID that is generated on the edge means that we need to have a way to share this ID with the platform. We could add an action that is called on startup to retrieve the ID but I think it would be better to wait until we have started the implementation of the new cloud architecture since I think the manager will be used for this step. I have added some text about this to the relevant work item.

skejserjensen (approved):

PR #84, opened 2023-03-26
Automatically finish unused in-memory buffers

skejserjensen:
This PR makes the `UncompressedDataManager` finish in-memory data buffers that are no longer used by tracking how many `RecordBatch`es has been received and which `RecordBatch` caused data points to be added to each in-memory data buffer. This reduces the amount of memory required when bulk-loading as there is no need to keep an open buffer for a time series after all of the data points for it has been loaded. At the end of `UncompressedDataManager::insert_data_points()` the buffers that have not been updated by the current batch is finished and thus compressed to release memory for additional uncompressed data points. For simplicity `UncompressedDataManager::insert_data_points()` currently perform the check after ingesting each `RecordBatch`, so as part of optimizing the storage engine a trigger for when to prune unused in-memory data buffers should be added.

### DIFF
--- a/crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
+++ b/crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
@@ -75,6 +75,8 @@ pub trait UncompressedDataBuffer: fmt::Debug + Sync + Send {
 pub(super) struct UncompressedInMemoryDataBuffer {
     /// Id that uniquely identifies the time series the buffer stores data points from.
     univariate_id: u64,
+    /// Index of the last batch that caused the buffer to be updated.
+    update_by_batch: u64,
     /// Error bound the buffer must be compressed within.
     error_bound: ErrorBound,
     /// Builder consisting of timestamps.
@@ -84,9 +86,10 @@ pub(super) struct UncompressedInMemoryDataBuffer {
 }
 
 impl UncompressedInMemoryDataBuffer {
-    pub(super) fn new(univariate_id: u64, error_bound: ErrorBound) -> Self {
+    pub(super) fn new(univariate_id: u64, current_batch: u64, error_bound: ErrorBound) -> Self {
         Self {
             univariate_id,
+            update_by_batch: current_batch,
             error_bound,
             timestamps: TimestampBuilder::with_capacity(UNCOMPRESSED_DATA_BUFFER_CAPACITY),
             values: ValueBuilder::with_capacity(UNCOMPRESSED_DATA_BUFFER_CAPACITY),
@@ -112,13 +115,21 @@ impl UncompressedInMemoryDataBuffer {
         self.timestamps.len()
     }
 
-    /// Add `timestamp` and `value` to the [`UncompressedInMemoryDataBuffer`].
-    pub(super) fn insert_data(&mut self, timestamp: Timestamp, value: Value) {
+    /// Return [`true`] if the [`UncompressedInMemoryDataBuffer`] was updated by the `batch`th batch
+    /// ingested by the current process.
+    pub(super) fn updated_by(&self, batch: u64) -> bool {
+        self.update_by_batch == batch
+    }
+
+    /// Add `timestamp` and `value` to the [`UncompressedInMemoryDataBuffer`] from the
+    /// `current_batch`th batch ingested by the current process.
+    pub(super) fn insert_data(&mut self, current_batch: u64, timestamp: Timestamp, value: Value) {
         debug_assert!(
             !self.is_full(),
             "Cannot insert data into full UncompressedInMemoryDataBuffer."
         );
 
+        self.update_by_batch = current_batch;
         self.timestamps.append_value(timestamp);
         self.values.append_value(value);
 
@@ -136,10 +147,11 @@ impl fmt::Debug for UncompressedInMemoryDataBuffer {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(
             f,
-            "UncompressedInMemoryDataBuffer({}, {}, {})",
+            "UncompressedInMemoryDataBuffer({}, {}, {}, {})",
             self.univariate_id,
             self.len(),
-            self.capacity()
+            self.capacity(),
+            self.update_by_batch,
         )
     }
 }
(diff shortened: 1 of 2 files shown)


crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -451,22 +518,25 @@ mod tests {
         // Spill an uncompressed buffer to disk.
         let temp_dir = tempfile::tempdir().unwrap();
         let error_bound = ErrorBound::try_new(0.0).unwrap();
-        let mut buffer = UncompressedInMemoryDataBuffer::new(UNIVARIATE_ID, error_bound);
-        buffer.insert_data(100, 10.0);
+        let mut buffer =
+            UncompressedInMemoryDataBuffer::new(UNIVARIATE_ID, CURRENT_BATCH, error_bound);
+        buffer.insert_data(100, 0, 10.0);
```
CGodiksen, line 523: Is it intentional that the batch index is 100 and the timestamp is 0 now. Previously the timestamp was 100 so it seems the batch index was inserted in the wrong position. Maybe also change it to use the `CURRENT_BATCH` constant.

crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -587,10 +660,68 @@ mod tests {
...
+                .active_uncompressed_data_buffers
+                .values()
+                .next()
+                .unwrap()
+                .len(),
+            3
+        );
+
+        // Insert using insert_data_points() to finish unused buffers.
+        let record_batch = RecordBatch::new_empty(model_table_metadata.schema.clone());
```
CGodiksen, line 709: Minor but maybe rename to `empty_record_batch` so it does not use the same name as the above record batch.

CGodiksen (changes requested):

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
```diff
@@ -112,13 +115,21 @@ impl UncompressedInMemoryDataBuffer {
         self.timestamps.len()
     }
 
-    /// Add `timestamp` and `value` to the [`UncompressedInMemoryDataBuffer`].
-    pub(super) fn insert_data(&mut self, timestamp: Timestamp, value: Value) {
+    /// Return [`true`] if the [`UncompressedInMemoryDataBuffer`] was updated by the `batch`th batch
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
```diff
@@ -112,13 +115,21 @@ impl UncompressedInMemoryDataBuffer {
...
+    /// ingested by the current process.
+    pub(super) fn updated_by(&self, batch: u64) -> bool {
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
```diff
@@ -112,13 +115,21 @@ impl UncompressedInMemoryDataBuffer {
...
+        self.update_by_batch == batch
+    }
+
+    /// Add `timestamp` and `value` to the [`UncompressedInMemoryDataBuffer`] from the
+    /// `current_batch`th batch ingested by the current process.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -42,6 +42,9 @@ pub(super) struct UncompressedDataManager {
     /// Path to the folder containing all uncompressed data managed by the
     /// [`StorageEngine`](crate::storage::StorageEngine).
     local_data_folder: PathBuf,
+    /// Counter incremented for each batch of data points ingested. The value is assigned to each
+    /// buffer that is updated by that call and used to flush buffers that is no longer updated.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -220,6 +224,16 @@ impl UncompressedDataManager {
             }
         }
 
+        // Unused buffers are only finished after ingesting a batch so that the buffers required for
```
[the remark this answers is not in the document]
skejserjensen: I have tried to simplify it.

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
```diff
@@ -36,6 +36,10 @@ use tracing::debug;
 
 use crate::storage::{StorageEngine, UNCOMPRESSED_DATA_BUFFER_CAPACITY, UNCOMPRESSED_DATA_FOLDER};
 
+/// Number of [`RecordBatches`](RecordBatch) that must be ingested without modifying an
+/// [`UncompressedInMemoryDataBuffer `] before it is considered unused and can be finished.
```
CGodiksen, line 40: It seems like there is an extra space in "[`UncompressedInMemoryDataBuffer `]"

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
```diff
@@ -389,27 +403,29 @@ mod tests {
     }
 
     #[test]
-    fn test_check_if_in_memory_data_buffer_is_updated_by_batch() {
+    fn test_check_if_in_memory_data_buffer_is_unused() {
+        // CURRENT_BATCH_INDEX is purposely not used to make all of the batch indexes visible in the test.
```
CGodiksen, line 407: Very minor but maybe use "indices" instead of "indexes" since it is used elsewhere.

CGodiksen (changes requested):

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
skejserjensen: Fixed.

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
skejserjensen: Deleted the comment since it is no longer relevant as `CURRENT_BATCH_INDEX` is used in the test.

CGodiksen (approved):

chrthomsen (approved):

PR #89, opened 2023-04-11
Support other ports than 9999 in the client

skejserjensen:
This PR adds support for other ports than 9999 to the client by allowing a non-default port to be specified using the syntax `host:port`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,16 @@ ModelarDB includes a command-line client in the form of `modelardb`. To interact
 modelardb
 ```
 
-If `modelardbd` is not running on the same host, the host `modelardb` should connect to must be specified. `modelardbd`'s Apache Arrow Flight interface accept requests on port `9999` so it is not necessary to specify a port:
+If `modelardbd` is not running on the same host, the host `modelardb` should connect to must be specified. `modelardbd`'s Apache Arrow Flight interface accept requests on port `9999` by default so it is generally not necessary to specify a port:
 
 ```shell
 modelardb 10.0.0.37
+```
+
+However, if `modelardbd` has been configured to use another port using the environment variable `MODELARDBD_PORT`, the same port must also be passed to the client:
+
+```shell
+modelardb 10.0.0.37:9998
 ```
 
 `modelardb` can also execute SQL statements from a file passed as a command-line argument:
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ ModelarDB includes a command-line client in the form of `modelardb`. To interact
 modelardb
 ```
 
-If `modelardbd` is not running on the same host, the host `modelardb` should connect to must be specified. `modelardbd`'s Apache Arrow Flight interface accept requests on port `9999` by default so it is generally not necessary to specify a port:
+If `modelardbd` is not running on the same host, the host `modelardb` should connect to must be specified. `modelardbd`'s Apache Arrow Flight interface accepts requests on port `9999` by default so it is generally not necessary to specify a port:
 
 ```shell
 modelardb 10.0.0.37
--- a/README.md
+++ b/README.md
@@ -225,14 +225,14 @@ used for purposes such as logging, where multiple crates provide similar functio
 
 ## Docker
 An environment that includes a local [MinIO](https://min.io/) instance and an edge node using the [MinIO](https://min.io/)
-instance as the remote object store, can be set up using [Docker](https://docs.docker.com/). Note that since 
+instance as the remote object store, can be set up using [Docker](https://docs.docker.com/). Note that since
 [Rust](https://www.rust-lang.org/) is a compiled language and a more dynamic `modelardbd` configuration might be needed,
-it is not recommended to use the [Docker](https://docs.docker.com/) environment during active development of `modelardbd`. 
+it is not recommended to use the [Docker](https://docs.docker.com/) environment during active development of `modelardbd`.
 It is however ideal to use during more static deployments or when developing components that utilize `modelardbd`.
 
-Downloading [Docker Desktop](https://docs.docker.com/desktop/) is recommended to make maintenance of the created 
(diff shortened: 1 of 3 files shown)


crates/modelardb_client/src/main.rs
```diff
@@ -92,29 +94,40 @@ fn main() -> Result<(), String> {
     .map_err(|error| format!("Cannot execute queries: {error}"))
 }
 
-/// Parse the command line arguments in `args` and return a pair with the address of the server to
-/// connect to and the file containing the queries to execute on the server. If one of the command
-/// line arguments is not provided it is replaced with [`None`] in the returned pair.
-fn parse_command_line_arguments(mut args: Args) -> (Option<String>, Option<String>) {
+/// Parse the command line arguments in `args` and return a triple with the host of the server to
+/// connect to, the port to connect to, and the file containing the queries to execute on the
+/// server. If one of these command line arguments is not provided it is replaced with [`None`]."
```
CGodiksen, line 99: I think there is an extra `"` at the end of this line.

crates/modelardb_client/src/main.rs
```diff
@@ -92,29 +94,40 @@ fn main() -> Result<(), String> {
...
             // Assumes all files contains queries.
             query_file = Some(arg);
+        } else if arg.contains(':') {
+            // Assumes anything with : is host:port.
+            let host_and_port = arg.splitn(2, ':').collect::<Vec<&str>>();
+            host = Some(host_and_port[0].to_owned());
+            port = Some(
+                host_and_port[1]
+                    .parse()
+                    .map_err(|_| "for host:port port must be from 1 to 65535.")
```
CGodiksen, line 121: Maybe remove `for host:port` from this error message.  Maybe also change it to `port must be between 1 and 65535.` or something similar.

crates/modelardb_server/src/main.rs
```diff
@@ -44,7 +44,10 @@ static ALLOC: snmalloc_rs::SnMalloc = snmalloc_rs::SnMalloc;
 
 /// The port of the Apache Arrow Flight Server. If the environment variable is not set, 9999 is used.
 pub static PORT: Lazy<u16> = Lazy::new(|| match env::var("MODELARDBD_PORT") {
-    Ok(port) => port.parse().unwrap(),
+    Ok(port) => port
+        .parse()
+        .map_err(|_| "MODELARDBD_PORT must be a port from 1 to 65535 if set")
```
CGodiksen, line 49: This error message is missing punctuation. Maybe also change it to `MODELARDBD_PORT must be between 1 and 65535.` or something similar.

CGodiksen (changes requested):

crates/modelardb_client/src/main.rs
skejserjensen: Fixed.

crates/modelardb_client/src/main.rs
skejserjensen: Fixed.

crates/modelardb_server/src/main.rs
skejserjensen: Fixed.

CGodiksen (approved):

crates/modelardb_client/src/main.rs
```diff
@@ -92,29 +94,40 @@ fn main() -> Result<(), String> {
...
             // Assumes all files contains queries.
             query_file = Some(arg);
+        } else if arg.contains(':') {
+            // Assumes anything with : is host:port.
+            let host_and_port = arg.splitn(2, ':').collect::<Vec<&str>>();
+            host = Some(host_and_port[0].to_owned());
+            port = Some(
+                host_and_port[1]
+                    .parse()
+                    .map_err(|_| "port must be between 1 and 65535.")
```
chrthomsen, line 121: would it make sense to require the port number to be between 1024 and 49151?

crates/modelardb_server/src/main.rs
```diff
@@ -44,7 +44,10 @@ static ALLOC: snmalloc_rs::SnMalloc = snmalloc_rs::SnMalloc;
 
 /// The port of the Apache Arrow Flight Server. If the environment variable is not set, 9999 is used.
 pub static PORT: Lazy<u16> = Lazy::new(|| match env::var("MODELARDBD_PORT") {
-    Ok(port) => port.parse().unwrap(),
+    Ok(port) => port
+        .parse()
+        .map_err(|_| "MODELARDBD_PORT must be between 1 and 65535.")
```
chrthomsen, line 49: 1024-49151?

README.md
```diff
@@ -101,12 +101,18 @@ ModelarDB includes a command-line client in the form of `modelardb`. To interact
 modelardb
 ```
 
-If `modelardbd` is not running on the same host, the host `modelardb` should connect to must be specified. `modelardbd`'s Apache Arrow Flight interface accept requests on port `9999` so it is not necessary to specify a port:
+If `modelardbd` is not running on the same host, the host `modelardb` should connect to must be specified. `modelardbd`'s Apache Arrow Flight interface accept requests on port `9999` by default so it is generally not necessary to specify a port:
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_client/src/main.rs
skejserjensen: I suggest we keep it between 1 and 65535 since we do not restrict the port number the user can use, so if the use has permission from the operating `modelardbd` runs fine on port 1 and 65535.

crates/modelardb_server/src/main.rs
skejserjensen: I suggest we keep it between 1 and 65535 since we do not restrict the port number the user can use, so if the use has permission from the operating `modelardbd` runs fine on port 1 and 65535.

crates/modelardb_client/src/main.rs
chrthomsen: ok

PR #90, opened 2023-04-16
Evaluate PMC-Mean and Swing before Gorilla

skejserjensen:
This PR modifies the compression in two ways. First, the size of the metadata is now included when computing the number of bytes per value a model requires, e.g., a single segment of length eight is better than two segments of length four due to the metadata. Second, Gorilla is no longer evaluated alongside PMC-Mean and Swing, instead it is used as a fallback when neither PMC-Mean nor Swing can create models that efficiently encode a sub-sequence of values. Another benefit of only using Gorilla for values is that PMC-Mean and Swing cannot compress, is that the length bound is no longer needed. Preliminary results indicates that this reduces the amount of storage required by 5-25% by using PMC-Mean and Swing for more values and decreases ingestion speed by 0-10% .

### DIFF
--- a/crates/modelardb_common/src/schemas.rs
+++ b/crates/modelardb_common/src/schemas.rs
@@ -50,8 +50,9 @@ pub static COMPRESSED_SCHEMA: Lazy<CompressedSchema> = Lazy::new(|| {
     ])))
 });
 
-/// Size of the metadata required for a compressed segment. Meaning that the size of `timestamps`
-/// and `values` are not included as they are binary and thus depends on which model is selected.
+/// Minimum size of the metadata required for a compressed segment. Meaning that the sizes of
+/// `timestamps` and `values` are not included as they are [`DataType::Binary`] and thus their size
+/// depend on which model is selected to represent the values for that compressed segment.
 pub static COMPRESSED_METADATA_SIZE_IN_BYTES: Lazy<usize> = Lazy::new(|| {
     COMPRESSED_SCHEMA
         .0
--- a/crates/modelardb_common/src/schemas.rs
+++ b/crates/modelardb_common/src/schemas.rs
@@ -50,6 +50,17 @@ pub static COMPRESSED_SCHEMA: Lazy<CompressedSchema> = Lazy::new(|| {
     ])))
 });
 
+/// Size of the metadata required for a compressed segment. Meaning that the size of `timestamps`
+/// and `values` are not included as they are binary and thus depends which model is selected.
+pub static COMPRESSED_METADATA_SIZE_IN_BYTES: Lazy<usize> = Lazy::new(|| {
+    COMPRESSED_SCHEMA
+        .0
+        .fields()
+        .iter()
+        .map(|field| field.data_type().primitive_width().unwrap_or(0))
+        .sum()
+});
+
 /// [`RecordBatch`](arrow::record_batch::RecordBatch) [`Schema`](arrow::datatypes::Schema) used for
 /// internally collected metrics.
 pub static METRIC_SCHEMA: Lazy<MetricSchema> = Lazy::new(|| {
--- a/crates/modelardb_common/src/schemas.rs
+++ b/crates/modelardb_common/src/schemas.rs
@@ -51,7 +51,7 @@ pub static COMPRESSED_SCHEMA: Lazy<CompressedSchema> = Lazy::new(|| {
 });
 
 /// Size of the metadata required for a compressed segment. Meaning that the size of `timestamps`
-/// and `values` are not included as they are binary and thus depends which model is selected.
+/// and `values` are not included as they are binary and thus depends on which model is selected.
 pub static COMPRESSED_METADATA_SIZE_IN_BYTES: Lazy<usize> = Lazy::new(|| {
     COMPRESSED_SCHEMA
         .0
--- a/crates/modelardb_compression/src/lib.rs
+++ b/crates/modelardb_compression/src/lib.rs
@@ -58,7 +58,7 @@ pub fn try_compress(
     }
 
     // Enough memory for end_index compressed segments are allocated to never require reallocation
-    // as one compressed segments is created per data point in the absolute worst case.
+    // as one compressed segment is created per data point in the absolute worst case.
     let end_index = uncompressed_timestamps.len();
     let mut compressed_record_batch_builder = CompressedSegmentBatchBuilder::new(end_index);
 
@@ -81,7 +81,7 @@ pub fn try_compress(
         // The selected model is only stored as part of a compressed segment if it uses less storage
         // space per value than the uncompressed values it represents.
         if selected_model.bytes_per_value <= models::VALUE_SIZE_IN_BYTES as f32 {
-            // Compress residual data points with gorilla and store them in a compressed segments.
+            // Compress residual data points with Gorilla and store them in a compressed segment.
             if residual_start_index != current_index {
                 compress_and_store_residual_value_range(
                     univariate_id,
@@ -93,7 +93,7 @@ pub fn try_compress(
                 )
             }
 
-            // Store the selected model and the corresponding timestamps in a compressed segments.
+            // Store the selected model and the corresponding timestamps in a compressed segment.
             store_selected_model_in_batch_builder(
                 univariate_id,
                 current_index,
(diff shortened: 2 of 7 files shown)


crates/modelardb_compression/src/lib.rs
```diff
@@ -32,62 +32,155 @@ use modelardb_common::types::{
...
     if uncompressed_timestamps.len() != uncompressed_values.len() {
         return Err(ModelarDbError::CompressionError(
             "Uncompressed timestamps and uncompressed values have different lengths.".to_owned(),
         ));
     }
 
-    // Enough memory for end_index elements are allocated to never require
-    // reallocation as one model is created per data point in the worst case.
+    // Enough memory for end_index compressed segments are allocated to never require reallocation
+    // as one compressed segments is created per data point in the absolute worst case.
```
CGodiksen, line 61: Maybe `as one compressed segment is created` instead.

crates/modelardb_compression/src/lib.rs
```diff
@@ -32,62 +32,155 @@ use modelardb_common::types::{
...
             error_bound,
         );
-        current_index = compressed_segment_builder.finish(&mut compressed_record_batch_builder);
+
+        let selected_model = compressed_model_builder.finish();
+
+        // The selected model is only stored as part of a compressed segment if it uses less storage
+        // space per value than the uncompressed values it represents.
+        if selected_model.bytes_per_value <= models::VALUE_SIZE_IN_BYTES as f32 {
+            // Compress residual data points with gorilla and store them in a compressed segments.
```
CGodiksen, line 84: Maybe `in a compressed segment.` instead.

crates/modelardb_compression/src/lib.rs
```diff
@@ -32,62 +32,155 @@ use modelardb_common::types::{
...
+                    univariate_id,
+                    residual_start_index,
+                    current_index - 1,
+                    uncompressed_timestamps,
+                    uncompressed_values,
+                    &mut compressed_record_batch_builder,
+                )
+            }
+
+            // Store the selected model and the corresponding timestamps in a compressed segments.
```
CGodiksen, line 96: Maybe `in a compressed segment.` instead.

crates/modelardb_compression/src/lib.rs
```diff
@@ -32,62 +32,155 @@ use modelardb_common::types::{
...
+            store_selected_model_in_batch_builder(
+                univariate_id,
+                current_index,
+                &selected_model,
+                uncompressed_timestamps,
+                &mut compressed_record_batch_builder,
+            );
+
+            // Update the indexes to specify that all data points until now has been compressed.
```
CGodiksen, line 105: Maybe use `indices` instead of `indexes`.

crates/modelardb_compression/src/lib.rs
```diff
@@ -32,62 +32,155 @@ use modelardb_common::types::{
...
+            current_index = selected_model.end_index + 1;
+            residual_start_index = current_index;
+        } else {
+            // The potentially lossy models could not efficiently encode the sub-sequence starting
+            // at current_index, the residual values will instead be compressed using gorilla.
```
CGodiksen, line 110: Since we are referring to the model `Gorilla`, maybe it should be capitalized. This might also need to be changed on line 84 and on line 115.

crates/modelardb_compression/src/lib.rs
```diff
@@ -309,60 +384,18 @@ impl<'a> CompressedSegmentBuilder<'a> {
...
-            || self.swing_could_fit_all
-            || self.gorilla.length < GORILLA_MAXIMUM_LENGTH
+        self.pmc_mean_could_fit_all || self.swing_could_fit_all
     }
 
-    /// Store the model that requires the smallest number of bits per value in
-    /// `compressed_record_batch_builder`. Returns the index of the first value
-    /// in `uncompressed_values` the selected model could not represent.
-    fn finish(self, compressed_record_batch_builder: &mut CompressedSegmentBatchBuilder) -> usize {
+    /// Return the model that requires the smallest number of bits per value if used for a segment.
```
CGodiksen, line 395: Maybe change to `the fewest number of bytes per value` since that unit is used elsewhere. I also don't fully understand the meaning of `if used for a segment` here.

crates/modelardb_compression/src/lib.rs
```diff
@@ -309,60 +384,18 @@ impl<'a> CompressedSegmentBuilder<'a> {
...
+    fn finish(self) -> SelectedModel {
         // The model that uses the fewest number of bytes per value is selected.
```
CGodiksen, line 397: Maybe remove this comment since this is already explained in the methods doc comment above.

CGodiksen (changes requested):

crates/modelardb_compression/src/models/gorilla.rs
```diff
@@ -119,7 +120,9 @@ impl Gorilla {
 
     /// Return the number of bytes currently used per data point on average.
     pub fn bytes_per_value(&self) -> f32 {
-        self.compressed_values.len() as f32 / self.length as f32
+        // Gorilla does not use metadata for encoding values, only the data in compressed_values.
+        (COMPRESSED_METADATA_SIZE_IN_BYTES.to_owned() as f32 + self.compressed_values.len() as f32)
```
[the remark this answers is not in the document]
skejserjensen: [Not as far as I am aware](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=35d2c7a18f3283b4301fe3c09d148ef4), the cast is now applied to the result.

crates/modelardb_compression/src/models/pmc_mean.rs
```diff
@@ -81,7 +82,8 @@ impl PMCMean {
 
     /// Return the number of bytes the current model uses per data point on average.
     pub fn bytes_per_value(&self) -> f32 {
-        models::VALUE_SIZE_IN_BYTES as f32 / self.length as f32
+        // No data is needed for PMC-Mean as the value can be read from min_value or max_value.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_compression/src/models/mod.rs
```diff
@@ -447,9 +437,35 @@ mod tests {
 
             pmc_mean_could_fit_all = pmc_mean_could_fit_all && pmc_mean.fit_value(value);
             swing_could_fit_all = swing_could_fit_all && swing.fit_data_point(timestamp, value);
-            gorilla.compress_value(value);
         }
-        SelectedModel::new(0, pmc_mean, swing, gorilla, uncompressed_values)
+        SelectedModel::new(0, pmc_mean, swing)
+    }
+
+    // Tests for compress_residual_value_range()
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_compression/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_compression/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_compression/src/lib.rs
skejserjensen: Fixed and removed the comment inside the method as it just repeated the doc comment.

crates/modelardb_compression/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_common/src/schemas.rs
```diff
@@ -50,6 +50,17 @@ pub static COMPRESSED_SCHEMA: Lazy<CompressedSchema> = Lazy::new(|| {
     ])))
 });
 
+/// Size of the metadata required for a compressed segment. Meaning that the size of `timestamps`
+/// and `values` are not included as they are binary and thus depends which model is selected.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

CGodiksen (approved):

chrthomsen (commented):
Looks good, I only have two very small comments

crates/modelardb_common/src/schemas.rs
```diff
@@ -50,6 +50,17 @@ pub static COMPRESSED_SCHEMA: Lazy<CompressedSchema> = Lazy::new(|| {
     ])))
 });
 
+/// Size of the metadata required for a compressed segment. Meaning that the size of `timestamps`
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

chrthomsen (approved):

PR #92, opened 2023-04-20
Support for Azure Blob Storage as initial remote data folder

CGodiksen:
This PR makes it possible to use Azure Blob Storage when starting the server with the `object_store_type://bucket_name` syntax for the argument. Since both S3 and Azure Blob Storage object stores now need to be validated when we start the server, the previous check function was also replaced by a generalized version that is used both when starting the server and when updating the remote object store using the action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,22 @@ Then, assuming a bucket named `wind-turbine` has been created through [MinIO's w
 modelardbd edge path_to_local_data_folder s3://wind-turbine
 ```
 
+`modelardbd` also supports using [Azure Blob Storage](https://azure.microsoft.com/en-us/products/storage/blobs/) 
+for the remote object store. To use [Azure Blob Storage](https://azure.microsoft.com/en-us/products/storage/blobs/), 
+set the following environment variables instead:
+
+```shell
+AZURE_STORAGE_ACCOUNT_NAME
+AZURE_STORAGE_ACCESS_KEY
+```
+
+`modelardbd` can then be run in edge mode with transfer of the ingested time series to the 
+[Azure Blob Storage](https://azure.microsoft.com/en-us/products/storage/blobs/) container `wind-turbine`:
+
+```shell
+modelardbd edge path_to_local_data_folder azureblobstorage://wind-turbine
+```
+
 To run `modelardbd` in cloud mode simply replace `edge` with `cloud` as shown below. Be aware that both a local data folder and an object store are required when `modelardbd` is run in cloud mode.
 
 ```shell
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The following commands are for Ubuntu Server. However, equivalent commands shoul
 5. Move `modelardbd` and `modelardb` from the `target` directory to any directory.
 
 ## Usage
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -81,6 +81,22 @@ Then, assuming a bucket named `wind-turbine` has been created through [MinIO's w
...
+for the remote object store. To use [Azure Blob Storage](https://azure.microsoft.com/en-us/products/storage/blobs/), 
+set the following environment variables instead:
+
+```shell
+AZURE_STORAGE_ACCOUNT_NAME
+AZURE_STORAGE_ACCESS_KEY
+```
+
+`modelardbd` can then be run in edge mode with transfer of the ingested time series to the 
+[Azure Blob Storage](https://azure.microsoft.com/en-us/products/storage/blobs/) container `wind-turbine`:
```
skejserjensen, line 94: Maybe explicitly if this assumes the container is created by the user like with MinIO or if it automatically created.

crates/modelardb_server/src/main.rs
```diff
@@ -110,9 +113,16 @@ fn main() -> Result<(), String> {
         Runtime::new().map_err(|error| format!("Unable to create a Tokio Runtime: {error}"))?,
     );
 
-    // Check if a remote data folder was provided and can be accessed. These checks are performed
+    // If a remote data folder was provided, check that it can be accessed. This check is performed
     // after parse_command_line_arguments() as the Tokio Runtime is required.
-    check_remote_data_folder(&runtime, &data_folders.remote_data_folder)?;
+    if let Some(remote_data_folder) = &data_folders.remote_data_folder {
+        // unwrap() is safe since if there is a remote data folder, there is always a third argument.
+        let remote_data_folder_type = arguments.get(2).unwrap().split_once("://").unwrap().0;
```
skejserjensen, line 120: To only parse the command line argument in `argument_to_remote_object_store()` and restrict `remote_data_folder_type` to the supported object stores, maybe create a enum with a variant for each object store which contain `Arc<dyn ObjectStore>`. Then after `validate_remote_data_folder()` has been run the object store can be extracted from the enum as we no longer need to know its type. Alternative a enum without any data could be created and returned from `argument_to_remote_object_store()` in a pair together with `Arc<dyn ObjectStore>`. Using either methods would remove the `&str` parameter from `validate_remote_data_folder()` which only accepts a very small number of arguments and remove the need for parsing `argument.get(2)` in two different places.

skejserjensen (changes requested):

crates/modelardb_server/src/main.rs
CGodiksen: I have changed the `validate_remote_data_folder()` function to accept an enum instead. However, i did not find a clean way to extract the enum value in `argument_to_remote_object_store` and pass it back. 

It would either require passing a runtime into the function to validate right after creating the object store or require that we pass the value back to main as you mention. Passing a runtime down breaks some tests so that is not ideal and passing the value back makes the code very messy. I did not find a way to cleanly extract the object store value from the enum when the enum variant is not known and passing back a pair only to avoid having to read the third argument twice seems overkill.

README.md
CGodiksen: Fixed.

crates/modelardb_server/src/main.rs
skejserjensen: I agree.

skejserjensen (approved):

chrthomsen (approved):

PR #93, opened 2023-04-21
Rewrite value if within error bound in Gorilla

skejserjensen:
This PR adds preliminary support for lossy compression to Gorilla by changing the current value to the previous value if possible within the error bound. Experiments showed that this can reduce the amount of storage required by up to 22%. I also tried rewriting the current value to be as close as possible numerically to the previous value if the current value cannot be reduced to the previous value within the error bound. However, this increased the amount of stored by up to  2.6% compared to only rewriting the current value if it is within the error bound of the previous value. It is probably possible to rewrite the current value to produce the smallest result when XOR'ed with the previous value, however, this would require a significant amount of work that probably cannot be reused if we switch to [Chimp](https://www.vldb.org/pvldb/vol15/p3058-liakos.pdf/) or [Elf](https://www.vldb.org/pvldb/vol16/p1763-li.pdf), so I suggest we simply rewrite the current value to the previous value if it is within the error bound for now.

### DIFF
--- a/crates/modelardb_compression/src/models/gorilla.rs
+++ b/crates/modelardb_compression/src/models/gorilla.rs
@@ -13,12 +13,12 @@
  * limitations under the License.
  */
 
-//! Implementation of the Gorilla model type which uses the lossless compression
-//! method for floating-point values proposed for the time series management
-//! system Gorilla in the [Gorilla paper]. As this compression method compresses
-//! the values of a time series segment using XOR and a variable length binary
-//! encoding, aggregates are computed by iterating over all values in the
-//! segment.
+//! Implementation of the Gorilla model type which uses the lossless compression method for
+//! floating-point values proposed for the time series management system Gorilla in the [Gorilla
+//! paper]. The compression method described in the paper has been extended with support for lossy
+//! compression by replacing values with the previous value if possible within the error bound. As
+//! this compression method compresses the values of a time series segment using XOR and a variable
+//! length binary encoding, aggregates are computed by iterating over all values in the segment.
 //!
 //! [Gorilla paper]: https://www.vldb.org/pvldb/vol8/p1816-teller.pdf
 
@@ -27,10 +27,13 @@ use modelardb_common::types::{Timestamp, UnivariateId, UnivariateIdBuilder, Valu
 
 use crate::models;
 use crate::models::bits::{BitReader, BitVecBuilder};
+use crate::models::ErrorBound;
 
 /// The state the Gorilla model type needs while compressing the values of a
 /// time series segment.
 pub struct Gorilla {
+    /// Maximum relative error for the value of each data point.
+    error_bound: ErrorBound,
     /// Last value compressed and added to `compressed_values`.
     last_value: Value,
     /// Number of leading zero bits for the last value that was compressed by
@@ -42,12 +45,13 @@ pub struct Gorilla {
     /// Values compressed using XOR and a variable length binary encoding.
     compressed_values: BitVecBuilder,
     /// The number of values stored in `compressed_values`.
-    pub length: usize, // TODO: use Gorilla as a fallback to remove pub.
+    length: usize,
 }
 
 impl Gorilla {
-    pub fn new() -> Self {
+    pub fn new(error_bound: ErrorBound) -> Self {
         Self {
+            error_bound,
             last_value: 0.0,
             last_leading_zero_bits: u8::MAX,
             last_trailing_zero_bits: 0,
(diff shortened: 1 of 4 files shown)


crates/modelardb_compression/src/models/gorilla.rs
```diff
@@ -13,12 +13,12 @@
...
 
-//! Implementation of the Gorilla model type which uses the lossless compression
-//! method for floating-point values proposed for the time series management
-//! system Gorilla in the [Gorilla paper]. As this compression method compresses
-//! the values of a time series segment using XOR and a variable length binary
-//! encoding, aggregates are computed by iterating over all values in the
-//! segment.
+//! Implementation of the Gorilla model type which uses the lossless compression method for
+//! floating-point values proposed for the time series management system Gorilla in the [Gorilla
+//! paper]. The compression method described in the paper has been extended with support for lossy
```
CGodiksen, line 18: This is not something that should change in this PR but I think we should make sure that we mention that `Gorilla` might be lossy in the documentation for the compression component. Potential users might always expect lossless compression when they hear `Gorilla` so it should be made very clear.

crates/modelardb_compression/src/models/mod.rs
```diff
@@ -49,10 +47,10 @@ pub(super) const VALUE_SIZE_IN_BYTES: u8 = mem::size_of::<Value>() as u8;
 /// Size of [`Value`] in bits.
 pub(super) const VALUE_SIZE_IN_BITS: u8 = 8 * VALUE_SIZE_IN_BYTES;
 
-/// General error bound that is guaranteed to not be negative, infinite, or NAN.
-/// For [`PMCMean`] and [`Swing`] the error bound is interpreted as a relative
-/// per value error bound in percentage, while [`Gorilla`] uses lossless
-/// compression.
+/// General error bound that is guaranteed to not be negative, infinite, or NAN. For [`PMCMean`],
+/// [`Swing`], and [`Gorilla`],  the error bound is interpreted as a relative per value error bound
```
CGodiksen, line 51: Minor but it seems there is one extra space after `[Gorilla],`.

crates/modelardb_compression/src/models/mod.rs
```diff
@@ -348,6 +365,59 @@ mod tests {
         assert!(ErrorBound::try_new(f32::NAN).is_err())
     }
 
+    // is_value_within_error_bound().
```
CGodiksen, line 368: Should be `Tests for is_value_within_error_bound().`.

crates/modelardb_compression/src/models/mod.rs
```diff
@@ -348,6 +365,59 @@ mod tests {
...
+    proptest! {
+    #[test]
+    fn test_same_value_is_always_within_error_bound(value in ProptestValue::ANY) {
+        prop_assert!(is_value_within_error_bound(ErrorBound::try_new(0.0).unwrap(), value, value));
+    }
+
+    fn test_other_value_is_never_within_error_bound_of_positive_infinity(value in ProptestValue::ANY) {
```
CGodiksen, line 375: It seems that these tests are missing the `#[test]` header. I am not sure the tests are being run currently.

crates/modelardb_compression/src/models/mod.rs
```diff
@@ -442,11 +512,17 @@ mod tests {
     }
 
     // Tests for compress_residual_value_range().
+    // TODO
```
CGodiksen, line 515: If this `TODO` should be kept maybe add some text to make it clear what should be done.

CGodiksen (changes requested):

crates/modelardb_compression/src/models/gorilla.rs
skejserjensen: Agreed, we should make it very clear that we used Gorilla's compression method as a baseline, but that we have extended with more functionality. According to `rg -i gorilla` we currently only mention `Gorilla` in the Rust files and I have tried to ensure that the comments make it clear that `Gorilla` is now lossy within an error bound.

crates/modelardb_compression/src/models/mod.rs
skejserjensen: Fixed.

crates/modelardb_compression/src/models/mod.rs
skejserjensen: Fixed.

crates/modelardb_compression/src/models/mod.rs
skejserjensen: Good catch, they are not. The tests are also wrong as I forgot to include `!` to specify that the special `f32` values should never be within the error bound of other values.

crates/modelardb_compression/src/models/mod.rs
skejserjensen: Deleted as it was a reminder that I should write tests that I forgot to delete.

CGodiksen (approved):

chrthomsen (approved):

PR #94, opened 2023-04-21
Restrict ErrorBound to 0-100 to improve errors

skejserjensen:
This PR explicitly restrict `ErrorBound` to percentage values from 0% to 100% in order to improve error messages.

### DIFF
--- a/crates/modelardb_compression/src/models/mod.rs
+++ b/crates/modelardb_compression/src/models/mod.rs
@@ -56,7 +56,7 @@ impl ErrorBound {
     /// Return [`ErrorBound`] if `percentage` is a value from 0.0% to 100.0%, otherwise
     /// [`CompressionError`](ModelarDbError::CompressionError) is returned.
     pub fn try_new(percentage: f32) -> Result<Self, ModelarDbError> {
-        if !(0.0..=100.0).contains(&percentage) || percentage.is_infinite() || percentage.is_nan() {
+        if !(0.0..=100.0).contains(&percentage) {
             Err(ModelarDbError::CompressionError(
                 "Error bound must be a value from 0.0% to 100.0%.".to_owned(),
             ))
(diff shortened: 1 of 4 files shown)


crates/modelardb_compression/src/models/mod.rs
```diff
@@ -49,28 +49,25 @@ pub(super) const VALUE_SIZE_IN_BYTES: u8 = mem::size_of::<Value>() as u8;
...
 
 impl ErrorBound {
-    /// Return [`ErrorBound`] if `error_bound` is a positive finite value,
-    /// otherwise [`CompressionError`](ModelarDbError::CompressionError).
-    pub fn try_new(error_bound: f32) -> Result<Self, ModelarDbError> {
-        if error_bound < 0.0 || error_bound.is_infinite() || error_bound.is_nan() {
+    /// Return [`ErrorBound`] if `percentage` is a value from 0.0% to 100.0%, otherwise
+    /// [`CompressionError`](ModelarDbError::CompressionError) is returned.
+    pub fn try_new(percentage: f32) -> Result<Self, ModelarDbError> {
+        if !(0.0..=100.0).contains(&percentage) || percentage.is_infinite() || percentage.is_nan() {
```
chrthomsen, line 61: Would the first check not also catch inf or NaN?

chrthomsen (commented):
Looks good. I just have one question

crates/modelardb_compression/src/models/mod.rs
skejserjensen: Should be and the [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cdc6623c27f475bbbcb43e6ba6c49e33) and the tests agrees.

CGodiksen (approved):

PR #97, opened 2023-05-01
Compute code coverage for all OSes

skejserjensen:
This PR uses GitHub Actions to compute code coverage for all crates in the repository using [LLVM's instrumentation-based coverage](https://doc.rust-lang.org/rustc/instrument-coverage.html) and then [grcov](https://github.com/mozilla/grcov) to produce a HTML report which is attached to the workflow run as an [artifact with a 90 days retention period](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts). A minor limitation of this approach is that the small number of Python unit test in repository is not included. GitHub does not seem to support [visualization code coverage like GitLab](https://docs.gitlab.com/ee/ci/testing/test_coverage_visualization.html), so it seems necessary to either [post comments with Markdown on PRs](https://github.com/marketplace/actions/code-coverage-summary), [host it on GitHub Pages](https://dev.to/mbarzeev/auto-publish-your-test-coverage-report-on-github-pages-35be), or [host it using a service like Codecov](https://about.codecov.io/for/open-source/) if the report needs to be more easily available in the future.

### DIFF
--- a/.github/workflows/cargo-build-lint-and-test-on-pr-and-push.yml
+++ b/.github/workflows/cargo-build-lint-and-test-on-pr-and-push.yml
@@ -22,8 +22,9 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTFLAGS: -D warnings
+  RUSTFLAGS: -C instrument-coverage -D warnings
   RUSTDOCFLAGS: -D warnings
+  LLVM_PROFILE_FILE: modelardb-%p-%m.profraw
 
 jobs:
   cargo_build_lint_and_test:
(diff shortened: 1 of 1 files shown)


skejserjensen:
The artifacts associated with the latest commit in this PR are now available [here](https://github.com/ModelarData/ModelarDB-RS/actions/runs/4853190650). The report is commuted for each OS as it was simpler than computing it for a single OS and future-proof the project in case OS-specific code is added. Be aware that the zip files contains multiple files and folder, thus it is advisable to extract them into an empty directory to make it simple to delete afterward.

CGodiksen (approved):

chrthomsen (approved):

PR #101, opened 2023-05-15
Refactor optimizer

skejserjensen:
This PR extends Apache Arrow DataFusion's optimizer with our optimizer rule instead of replacing the existing rules with our, refactors the optimizer module, documents the optimizer module, and adds tests to the optimizer module.

PR #102, opened 2023-05-17
Fix the operator metadata used for planning

skejserjensen:
As Apache Arrow DataFusion's optimizer rules has been re-enabled by #101, this PR updates the metadata returned by each  [`ExecutionPlan`](https://docs.rs/datafusion/latest/datafusion/physical_plan/trait.ExecutionPlan.html) added to Apache Arrow DataFusion in `modelardbd`. This is to ensure Apache Arrow DataFusion's optimizer knows which optimizations are possible, and more importantly, safe. For example, n * [`ParquetExec` -> `GridExec` -> `PartitionExec`] -> `SortedJoinExec` is not safe as far as I can tell since the documentation for [`ExecutionPlan::required_input_ordering()`](https://docs.rs/datafusion/latest/datafusion/physical_plan/trait.ExecutionPlan.html#method.required_input_ordering) states that it _"Specifies the ordering requirements for all of the children. For each child, it’s the local ordering requirement within each partition rather than the global ordering"_ and the documentation for [`FileScanConfig::file_groups`](https://docs.rs/datafusion/latest/datafusion/physical_plan/file_format/struct.FileScanConfig.html#structfield.file_groups) (`FileScanConfig` is used to configure [`ParquetExec`](https://docs.rs/datafusion/latest/datafusion/physical_plan/file_format/struct.ParquetExec.html)) states that _"Apache Arrow DataFusion may attempt to read each partition of files concurrently, however files within a partition will be read sequentially, one after the next."_ so it does not seem possible to use multiple partitions while also ensuring that the batches of data points received by `SortedJoinExec` all has the same sort order.

### DIFF
--- a/crates/modelardb_server/src/query/mod.rs
+++ b/crates/modelardb_server/src/query/mod.rs
@@ -77,7 +77,7 @@ pub static QUERY_ORDER_SEGMENT: Lazy<Vec<PhysicalSortExpr>> = Lazy::new(|| {
             options: sort_options,
         },
         PhysicalSortExpr {
-            expr: Arc::new(Column::new("start_time", 1)),
+            expr: Arc::new(Column::new("start_time", 2)),
             options: sort_options,
         },
     ]
--- a/crates/modelardb_server/src/query/mod.rs
+++ b/crates/modelardb_server/src/query/mod.rs
@@ -27,6 +27,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use datafusion::arrow::compute::SortOptions;
 use datafusion::arrow::datatypes::{
     ArrowPrimitiveType, DataType, Field, Schema, SchemaRef, TimeUnit,
 };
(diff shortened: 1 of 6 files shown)


crates/modelardb_server/src/query/mod.rs
```diff
@@ -55,6 +58,56 @@ use crate::storage;
...
+        },
+        PhysicalSortExpr {
+            expr: Arc::new(Column::new("start_time", 2)),
+            options: sort_options,
+        },
+    ]
+});
+
+/// The global sort order [`GridExec`] guarantees for the data points it produces and that
+/// [`SortedJoinExec`] requires for the data points its receives as its input. It is guaranteed by
```
CGodiksen, line 87: Maybe `it receives` instead.

crates/modelardb_server/src/query/sorted_join_exec.rs
```diff
@@ -165,6 +170,21 @@ impl ExecutionPlan for SortedJoinExec {
         }
     }
 
+    /// Specify that [`SortedJoinStream`] requires one partition for each input as it assumes that
+    /// the global sort order are the same for all inputs and Apache Arrow DataFusion only
```
CGodiksen, line 174: Maye `is the same` instead.

CGodiksen (changes requested):

crates/modelardb_server/src/query/mod.rs
skejserjensen: Fixed.

crates/modelardb_server/src/query/sorted_join_exec.rs
skejserjensen: Fixed.

CGodiksen (approved):

chrthomsen (approved):

PR #104, opened 2023-05-18
Improve runtime errors for generated columns

skejserjensen:
This PR adds the column name and row to the error message that are returned if the values for a generated column cannot be generated, e.g., if the [PhysicalExpr](https://docs.rs/datafusion-physical-expr/latest/datafusion_physical_expr/trait.PhysicalExpr.html) divides by zero. Unfortunately, as far as I can tell, it is not possible to extract which row caused the error from [PhysicalExpr](https://docs.rs/datafusion-physical-expr/latest/datafusion_physical_expr/trait.PhysicalExpr.html), [DataFusionError](https://docs.rs/datafusion/latest/datafusion/error/enum.DataFusionError.html), or [ArrowError](https://docs.rs/arrow-schema/38.0.0/arrow_schema/enum.ArrowError.html). So the method `GeneratedAsStream::failing_row()` has been added which tries to evaluate the failed [PhysicalExpr](https://docs.rs/datafusion-physical-expr/latest/datafusion_physical_expr/trait.PhysicalExpr.html) for each row in the [`RecordBatch`](https://docs.rs/arrow/latest/arrow/record_batch/struct.RecordBatch.html). Since [`RecordBatch::slice()`](https://docs.rs/arrow/latest/arrow/record_batch/struct.RecordBatch.html#method.slice) simply increments atomic reference counts (the documentation states that [`Array::slice()`](https://github.com/apache/arrow-rs/blob/master/arrow-array/src/array/mod.rs#L131) is always zero-copy) and only allocates a single short [`Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html), the method is fast enough to execute in the case of a rare error that terminates the query. A small scale performance test with a table producing 304 [`RecordBatches`](https://docs.rs/arrow/latest/arrow/record_batch/struct.RecordBatch.html) containing 8912 rows each, showed that the method executed in 8ms - 10ms for 276 of the [`RecordBatches`](https://docs.rs/arrow/latest/arrow/record_batch/struct.RecordBatch.html) and 10ms - 20ms for 28 of the [`RecordBatches`](https://docs.rs/arrow/latest/arrow/record_batch/struct.RecordBatch.html) on a virtual machine with 8 vCPU and 16 GiB of memory. [`chrono`](https://crates.io/crates/chrono/0.4.24) is used for formatting the timestamps if possible as it is already a dependency of [`Apache Arrow DataFusion`](https://crates.io/crates/datafusion/24.0.0/dependencies) and is also used in the [Rust implementation of Apache Arrow to format timestamps](https://github.com/apache/arrow-rs/blob/master/arrow-cast/src/display.rs#L451).

### DIFF
--- a/crates/modelardb_server/src/query/generated_as_exec.rs
+++ b/crates/modelardb_server/src/query/generated_as_exec.rs
@@ -199,9 +199,10 @@ impl GeneratedAsStream {
         }
     }
 
-    /// Format and return the first row in `batch` that causes `physical_expr` to fail, if
-    /// `physical_expr` never fails or if `batch` is from a normal table, [`None`] is returned.
-    fn failing_row(batch: &RecordBatch, physical_expr: &Arc<dyn PhysicalExpr>) -> Option<String> {
+    /// Format and return the first row in `batch` that causes `physical_expr` to fail when
+    /// generating values. If `physical_expr` never fails the string Unknown Row is returned, and a
+    /// [`DataFusionError`] is returned if `batch` is from a normal table.
+    fn failing_row(batch: &RecordBatch, physical_expr: &Arc<dyn PhysicalExpr>) -> Result<String> {
         for row_index in 0..batch.num_rows() {
             if physical_expr.evaluate(&batch.slice(row_index, 1)).is_err() {
                 let schema = batch.schema();
@@ -229,16 +230,16 @@ impl GeneratedAsStream {
                         formatted_values.push(format!("{name}: {}", tags.value(row_index)));
                     } else {
                         // The method has been called for a table with unsupported column types.
-                        return None;
+                        return Err(DataFusionError::Internal("Not a model table.".to_owned()));
                     }
                 }
 
-                return Some(formatted_values.join(", "));
+                return Ok(formatted_values.join(", "));
             }
         }
 
-        // physical_expr never failed for any of the rows in batch.
-        None
+        // physical_expr never failed for any of the rows in batch, e.g., if random values are used.
+        Ok("Unknown Row".to_owned())
     }
 }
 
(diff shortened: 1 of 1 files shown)


CGodiksen (approved):

crates/modelardb_server/src/query/generated_as_exec.rs
```diff
@@ -195,6 +198,48 @@ impl GeneratedAsStream {
             baseline_metrics,
         }
     }
+
+    /// Format and return the first row in `batch` that causes `physical_expr` to fail, if
+    /// `physical_expr` never fails or if `batch` is from a normal table, [`None`] is returned.
+    fn failing_row(batch: &RecordBatch, physical_expr: &Arc<dyn PhysicalExpr>) -> Option<String> {
+        for row_index in 0..batch.num_rows() {
+            if physical_expr.evaluate(&batch.slice(row_index, 1)).is_err() {
```
chrthomsen, line 206: Can it happen that when the user SELECTs from a generated column it fails, but when we try here it does not? If RANDOM(), .e.g., is used in the expression that could potentially (although, unlikely) happen

crates/modelardb_server/src/query/generated_as_exec.rs
```diff
@@ -227,8 +272,17 @@ impl Stream for GeneratedAsStream {
...
+                        let column_name = self.schema.field(column_to_generate.index).name();
+
                         // unwrap() is safe as it is only executed if a column was not generated.
-                        return Poll::Ready(Some(Err(maybe_generated_column.err().unwrap())));
+                        let physical_expr = &column_to_generate.physical_expr;
+                        let failing_row = Self::failing_row(&batch, physical_expr).unwrap();
+                        let cause = maybe_generated_column.err().unwrap();
+
+                        let error = format!(
+                            "Failed to create '{column_name}' for {{{failing_row}}} due to: {cause}"
```
chrthomsen, line 283: "create" -> "compute" ?

skejserjensen: Fixed.

crates/modelardb_server/src/query/generated_as_exec.rs
skejserjensen: Yes that should be possible since we execute the physical expression again. One solution is to do like MySQL and just reject functions that can cause such problems: `ERROR 3763 (HY000): Expression of generated column 'generated' contains a disallowed function: rand.` However, this would require us to continuously maintain a list of functions that Apache Arrow DataFusion implements which can cause such problems and limits what users can do. So instead I have updated the function to return an error if the table is not a model table (i.e., we have a bug in the planner) and otherwise return the row as a string or the string "Unknown Row".

chrthomsen (approved):

PR #111, opened 2023-06-05
Fix wrong comparison when encoding of Swing

skejserjensen:
This PR corrects a typo in `CompressedSegmentBuilder::update_values_for_swing()` by replacing `residuals_min_value < self.min_value && self.max_value < residuals_min_value` with `residuals_min_value < self.min_value && self.max_value < residuals_max_value` such that segments containing a model of type Swing is stored correctly even if the residuals includes both the segment's minimum and maximum value. To decrease that chance that more issues in the code the encodes and decodes models exists, the amount of unit tests for this code has also been significantly increased.

### DIFF
--- a/crates/modelardb_compression/src/types.rs
+++ b/crates/modelardb_compression/src/types.rs
@@ -559,25 +559,49 @@ mod tests {
     #[test]
     fn test_encoding_decoding_for_pmc_mean_without_residuals() {
         let uncompressed_values = ValueArray::from(vec![10.0, 10.0, 10.0, 10.0, 10.0]);
-        assert_encoding_decoding_for_pmc_mean(uncompressed_values, 4, 10.0, 0, 10.0, 10.0, 0);
+        assert_encoding_and_decoding_are_valid_for_pmc_mean(
+            uncompressed_values,
+            4,
+            10.0,
+            0,
+            10.0,
+            10.0,
+            0,
+        );
     }
 
     #[test]
     fn test_encoding_decoding_for_pmc_mean_with_residuals_with_new_mininum() {
         let uncompressed_values = ValueArray::from(vec![10.0, 10.0, 10.0, 10.0, Value::MIN]);
-        assert_encoding_decoding_for_pmc_mean(uncompressed_values, 3, 10.0, 0, Value::MIN, 10.0, 1);
+        assert_encoding_and_decoding_are_valid_for_pmc_mean(
+            uncompressed_values,
+            3,
+            10.0,
+            0,
+            Value::MIN,
+            10.0,
+            1,
+        );
     }
 
     #[test]
     fn test_encoding_decoding_for_pmc_mean_with_residuals_with_new_maximum() {
         let uncompressed_values = ValueArray::from(vec![10.0, 10.0, 10.0, 10.0, Value::MAX]);
-        assert_encoding_decoding_for_pmc_mean(uncompressed_values, 3, 10.0, 0, 10.0, Value::MAX, 0);
+        assert_encoding_and_decoding_are_valid_for_pmc_mean(
+            uncompressed_values,
+            3,
+            10.0,
+            0,
+            10.0,
+            Value::MAX,
+            0,
+        );
     }
 
     #[test]
     fn test_encoding_decoding_for_pmc_mean_with_residuals_with_new_minimum_and_maximum() {
         let uncompressed_values = ValueArray::from(vec![10.0, 10.0, 10.0, Value::MIN, Value::MAX]);
-        assert_encoding_decoding_for_pmc_mean(
+        assert_encoding_and_decoding_are_valid_for_pmc_mean(
             uncompressed_values,
             2,
             10.0,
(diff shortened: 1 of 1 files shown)


chrthomsen (approved):

crates/modelardb_compression/src/types.rs
```diff
@@ -555,58 +557,263 @@ mod tests {
...
-        assert_eq!(uncompressed_timestamps.len() - 1, model.end_index);
-        assert_eq!(10.0, model.min_value);
-        assert_eq!(10.0, model.max_value);
-        assert_eq!(0, model.values.len());
+        assert_eq!(expected_model_end_index, model.end_index);
+        assert_eq!(expected_model_min_max_value, model.min_value);
+        assert_eq!(expected_model_min_max_value, model.max_value);
+        assert_eq!(expected_model_values_length, model.values.len());
+
+        // Create a segment that represents its values using a model of type Swing and its residuals
```
CGodiksen, line 617: Should this be `using a model of type PMC-Mean` instead?

crates/modelardb_compression/src/types.rs
```diff
@@ -555,58 +557,263 @@ mod tests {
...
+            30.0,
+            50.0,
+            1,
+            Value::MIN,
+            Value::MAX,
+            8,
+        );
+    }
+
+    fn test_encoding_decoding_for_swing(
```
CGodiksen, line 753: This is not extremely important since they are just test functions but there is a lot of duplicated code between `test_encoding_decoding_for_swing` and `test_encoding_decoding_for_pmc_mean`. Maybe create combine them or create separate functions to for example create the segment to avoid as much duplicated code.

crates/modelardb_compression/src/types.rs
```diff
@@ -555,58 +557,263 @@ mod tests {
...
+            2,
+            10.0,
+            0,
+            Value::MIN,
+            Value::MAX,
+            4,
+        );
+    }
+
+    fn test_encoding_decoding_for_pmc_mean(
```
CGodiksen, line 591: I am not sure this is a convention we currently use consistently but maybe we should attempt to avoid naming test utility functions such as this with the "test_" prefix and only use this prefix for actual tests. Utility functions that use `assert!` could maybe be named `assert_` instead.

crates/modelardb_compression/src/types.rs
CGodiksen, line 753: I am not sure this is a convention we currently use consistently but maybe we should attempt to avoid naming test utility functions such as this with the "test_" prefix and only use this prefix for actual tests. Utility functions that use `assert!` could maybe be named `assert_` instead.

CGodiksen (changes requested):

crates/modelardb_compression/src/types.rs
skejserjensen: Fixed.

crates/modelardb_compression/src/types.rs
skejserjensen: Fixed.

skejserjensen: Fixed.

crates/modelardb_compression/src/types.rs
```diff
@@ -750,7 +726,7 @@ mod tests {
         );
     }
 
-    fn test_encoding_decoding_for_swing(
+    fn assert_encoding_decoding_for_swing(
```
CGodiksen, line 729: Maybe rename to something like `assert_encoding_decoding_valid_for_swing`.

crates/modelardb_compression/src/types.rs
```diff
@@ -760,28 +736,72 @@ mod tests {
...
+                segment_min_value,
+                segment_max_value,
+                &segment_values,
+            );
+
+        assert_eq!(model_first_value, segment_first_value);
+        assert_eq!(model_last_value, segment_last_value);
+    }
+
+    fn create_segment_assert_segment_and_return_model_values(
```
CGodiksen, line 771: Seems odd to mention the return value in the function name like this, maybe rename to something like `create_and_assert_expected_segment`.

CGodiksen (changes requested):

crates/modelardb_compression/src/types.rs
skejserjensen: Fixed.

crates/modelardb_compression/src/types.rs
```diff
@@ -588,7 +588,7 @@ mod tests {
         );
     }
 
-    fn test_encoding_decoding_for_pmc_mean(
+    fn assert_encoding_decoding_for_pmc_mean(
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_compression/src/types.rs
skejserjensen: Fixed.

CGodiksen (approved):

PR #115, opened 2023-06-13
Support storing metadata for files in database

skejserjensen:
This PR adds support for storing metadata about the Apache Parquet files that the storage engine creates for storing the compressed segments. By storing the metadata in the metadata database instead of in the file names, multiple file operations can be performed atomically as the files are first accessible when their metadata has been written to the metadata database. To significantly simplify retrieving files based on their time and value range, the special floating-point values are rewritten to their closets normal floating value by `rewrite_special_value_to_normal_value(value: Value)` before they are stored in the metadata database. While this reduces the metadata managers ability to prune away files slightly, e..g, `NaN` and `f32::MAX` becomes the same value, it seems to be uncommon for sensors that monitor physical entities to report values near `f32::MIN` or `f32::MAX` and [`ParquetExec`](https://docs.rs/datafusion/26.0.0/datafusion/physical_plan/file_format/struct.ParquetExec.html) prunes irrelevant data in each  file if it is passed a `predicate`. Thus, in practice, rewriting the special values to normal values should significantly simplify retrieving information about the the relevant files from the metadata database as simple SQL range queries can be used and the reduction in pruning power should only very rarely have an effect.

### DIFF
--- a/crates/modelardb_server/src/metadata/mod.rs
+++ b/crates/modelardb_server/src/metadata/mod.rs
@@ -41,12 +41,15 @@ use datafusion::execution::options::ParquetReadOptions;
 use futures::TryStreamExt;
 use log::LevelFilter;
 use modelardb_common::errors::ModelarDbError;
-use modelardb_common::types::UnivariateId;
+use modelardb_common::types::{Timestamp, UnivariateId, Value};
 use modelardb_compression::ErrorBound;
+use sqlx::database::HasArguments;
 use sqlx::error::Error;
+use sqlx::query::Query;
 use sqlx::sqlite::{SqliteConnectOptions, SqliteRow};
-use sqlx::{ConnectOptions, Executor, Result, Row, SqlitePool};
+use sqlx::{ConnectOptions, Executor, Result, Row, Sqlite, SqlitePool};
 use tracing::{error, info, warn};
+use uuid::Uuid;
 
 use crate::metadata::model_table_metadata::ModelTableMetadata;
 use crate::parser;
@@ -58,6 +61,39 @@ use self::model_table_metadata::GeneratedColumn;
 
 /// Name used for the file containing the SQLite database storing the metadata.
 pub const METADATA_DATABASE_NAME: &str = "metadata.sqlite3";
+
+/// Metadata about a file tracked by [`MetadataManager`] which contains compressed segments.
+#[derive(Debug, Clone)]
+pub struct CompressedFile {
+    /// Name of the file.
+    name: Uuid,
+    /// Timestamp of the first data point in the file.
+    start_time: Timestamp,
+    /// Timestamp of the last data point in the file.
+    end_time: Timestamp,
+    /// Value of the data point with the smallest value in the file.
+    min_value: Value,
+    /// Value of the data point with the largest value in the file.
+    max_value: Value,
+}
+
+impl CompressedFile {
+    pub fn new(
+        name: Uuid,
+        start_time: Timestamp,
+        end_time: Timestamp,
+        min_value: Value,
+        max_value: Value,
+    ) -> Self {
+        Self {
+            name,
+            start_time,
+            end_time,
+            min_value,
+            max_value,
+        }
+    }
+}
 
 /// Store's the system's configuration and the metadata required for reading from and writing to the
 /// tables and model tables. The data that needs to be persisted are stored in the metadata
@@ -92,7 +128,7 @@ impl MetadataManager {
         let mut options = SqliteConnectOptions::new()
             .filename(local_data_folder.join(METADATA_DATABASE_NAME))
             .create_if_missing(true);
- 
+
         options.log_statements(LevelFilter::Debug);
 
         // Create the metadata manager with the default values.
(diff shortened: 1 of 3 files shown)


crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -460,6 +499,244 @@ impl MetadataManager {
...
+        let mut file_names = vec![];
+        while let Some(row) = rows.try_next().await? {
+            file_names.push(row.try_get(0)?);
+        }
+
+        Ok(file_names)
+    }
+
+    /// Rewrite the special values [`Value`] (Negative Infinity, Infinity, and NaN) can represent to
+    /// the closets normal values in [`Value`] to simplify storing and querying them through SQLite.
```
CGodiksen, line 728: is this specific to SQLite or does it also benefit the other supported databases? If not general then maybe we should start to compile a list of what needs to change when moving to support more databases and add this to the list.

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -560,6 +836,18 @@ impl MetadataManager {
             )
             .await?;
 
+        // Create a table_name_compressed_files SQLite table to save the name of the table's files.
```
CGodiksen, line 839: Maybe add the creation of this table to the function doc comment.

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -1105,6 +1406,494 @@ mod tests {
...
+                "table_model",
+                &[compressed_file]
+            )
+            .await
+            .is_err()
+        );
+    }
+
+    #[tokio::test]
+    async fn test_save_compressed_file_with_swaped_timestamps_for_column_in_existing_model_table() {
```
CGodiksen, line 1436: Should it be `swapped` instead? Maybe just replace it with `invalid`.

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -1105,6 +1406,494 @@ mod tests {
...
+                "model_table",
+                &[compressed_file]
+            )
+            .await
+            .is_err()
+        );
+    }
+
+    #[tokio::test]
+    async fn test_save_compressed_file_with_swaped_values_for_column_in_existing_model_table() {
```
CGodiksen, line 1449: Should it be `swapped` instead? Maybe just replace it with `invalid`.

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -1105,6 +1406,494 @@ mod tests {
...
+            create_metadata_manager_with_named_model_table_and_save_files(
+                "model_table",
+                &[compressed_file]
+            )
+            .await
+            .is_err()
+        );
+    }
+
+    async fn create_metadata_manager_with_named_model_table_and_save_files(
```
CGodiksen, line 1461: Maybe move this function down below its last use. I am not sure how consistent we are with this but it seems like something we have tried to consistently do.

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -1105,6 +1406,494 @@ mod tests {
...
+                    compressed_file,
+                )
+                .await?;
+        }
+
+        Ok(metadata_manager)
+    }
+
+    #[tokio::test]
+    async fn test_replace_compressed_files_for_column_in_model_table_with_files_to_delete() {
```
CGodiksen, line 1482: I am not sure i understand this test name. It makes it sound like the compressed files are replaced with the files to delete. Does it not just test that the files are replaced with `None` correctly?

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -1105,6 +1406,494 @@ mod tests {
...
+
+        assert_eq!(4, returned_files.len());
+        assert_eq!(compressed_files[3].name, returned_files[0]);
+        assert_eq!(compressed_files[4].name, returned_files[1]);
+        assert_eq!(compressed_files[5].name, returned_files[2]);
+        assert_eq!(compressed_files[6].name, returned_files[3]);
+    }
+
+    #[tokio::test]
+    async fn test_replace_compressed_files_for_column_in_model_table_with_files_to_replace() {
```
CGodiksen, line 1505: I am not sure i understand this test name. It makes it sound like the compressed files are replaced with the files to replace. Maybe rename it to something like `test_replace_compressed_files_for_column_in_model_table_with_replacement_file`

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -1105,6 +1406,494 @@ mod tests {
...
+            end_time,
+            min_value,
+            max_value,
+        )
+        .await;
+
+        (compressed_files, returned_files)
+    }
+
+    async fn create_metadata_manager_with_model_table_save_files_and_get_files(
```
CGodiksen, line 1870: I do not think that this is a problem that can easily be fixed but I do think that the tests are harder to understand when we rely heavily on these very specific undocumented utility functions like this one.

CGodiksen: It might also make the tests harder to understand when we include these functions in the middle of tests. It might be easier to get an overview of what utility functions are used and whether some can be combined/refactored if they were defined together after the tests. 

Since we have used the convention of putting test utility functions in the middle of the tests elsewhere, this might be something that should be discussed before a decision is made.

CGodiksen (changes requested):

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -460,6 +499,244 @@ impl MetadataManager {
...
+                )),
+            )));
+        };
+
+        Ok(())
+    }
+
+    /// Create a [`Query`] that, when executed, stores `compressed_file` in the metadata database
+    /// for the column at `query_schema_index` using `insert_statement`.
+    fn create_query_that_save_compressed_file_for_column_in_model_table<'a>(
```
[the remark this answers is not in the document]
skejserjensen: I do not have strong opinions on "that" function names, so I changed the name to your first suggestion.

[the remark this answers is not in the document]
skejserjensen: I am not aware of any recommendations regarding naming at the Rust level and a quick search did not bring up anything. The main reason for my use of very long  is names that other developers on the JVM-based version of ModelarDB found the names too terse, e.g., for a small example see [SparkStorage](https://github.com/ModelarData/ModelarDB/blob/master/src/main/scala/dk/aau/modelardb/engines/spark/SparkStorage.scala), so it is also fine for me to use simpler names for methods added in this PR such as `MetadataManager::save_compressed_files()`, `MetadataManager::replace_compressed_files()`, and `create_insert_compressed_file_query` if you prefer?

crates/modelardb_server/src/metadata/mod.rs
```diff
@@ -460,6 +499,244 @@ impl MetadataManager {
...
+    /// * A model table with `model_table_name` does not exist.
+    pub async fn compressed_files_for_column_in_model_table(
+        &self,
+        model_table_name: &str,
+        query_schema_index: usize,
+        start_time: Option<Timestamp>,
+        end_time: Option<Timestamp>,
+        min_value: Option<Value>,
+        max_value: Option<Value>,
+    ) -> Result<Vec<Uuid>> {
```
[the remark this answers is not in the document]
skejserjensen: Fixed by updating the doc comment.

crates/modelardb_server/src/metadata/mod.rs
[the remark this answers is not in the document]
CGodiksen: I think i prefer names like the ones you mentioned. What do you think @chrthomsen?

crates/modelardb_server/src/metadata/mod.rs
skejserjensen: It also benefits the other support RDBMSs, so I have changed `SQLite` to `metadata database`.

crates/modelardb_server/src/metadata/mod.rs
skejserjensen: Fixed.

crates/modelardb_server/src/metadata/mod.rs
skejserjensen: Fixed by changing it to invalid.

crates/modelardb_server/src/metadata/mod.rs
skejserjensen: Fixed.

crates/modelardb_server/src/metadata/mod.rs
skejserjensen: Fixed by renaming it to `....with_nothing()` to keep the naming of the tests for that function consistent.

crates/modelardb_server/src/metadata/mod.rs
skejserjensen: Fixed using suggested name.

crates/modelardb_server/src/metadata/mod.rs
skejserjensen: I agree and have added a task.

CGodiksen (approved):

PR #116, opened 2023-06-15
Update the installation guide for Windows

skejserjensen:
This PR updates the installation instructions to make it clear that Microsoft Visual C++ (MSVC) and Python 3 is required to build all of the crates in the project on Microsoft Windows. Since the MSVC components the `rustup` book states should be installed before the MSVC-based Rust toolchain can be used on Microsoft Windows are sufficient to compile all creates in the proejct, our simple installation instructions have been replaced with a link to the much more detailed instructions in the `rustup` book.

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -16,7 +16,7 @@ The following commands are for Ubuntu Server. However, equivalent commands shoul
 1. Install [cURL](https://curl.se/) as the *root* user: `pkg install curl`
 
 ### Windows
-1. Install the Microsoft Visual C++ Prerequisites for Rust and Python 3:
+1. Install the latest versions of the Microsoft Visual C++ Prerequisites for Rust and Python 3:
    - Microsoft Visual C++ Prerequisites for Rust: see [The rustup book](https://rust-lang.github.io/rustup/installation/windows-msvc.html).
    - Python 3: install [Python 3](https://www.python.org/) and select _"Add python.exe to PATH"_.
 
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -16,9 +16,9 @@ The following commands are for Ubuntu Server. However, equivalent commands shoul
 1. Install [cURL](https://curl.se/) as the *root* user: `pkg install curl`
 
 ### Windows
-1. Install a supported version of [Visual Studio](https://visualstudio.microsoft.com/vs/older-downloads/) with Visual C++:
-   - [Visual Studio 2019](https://visualstudio.microsoft.com/vs/older-downloads/#visual-studio-2019-and-other-products) ([Supported](https://github.com/microsoft/snmalloc/blob/main/docs/BUILDING.md#building-on-windows))
-   - [Visual Studio 2022](https://visualstudio.microsoft.com/vs/) ([Supported](https://github.com/microsoft/snmalloc/blob/main/docs/BUILDING.md#building-on-windows))
+1. Install the Microsoft Visual C++ Prerequisites for Rust and Python 3:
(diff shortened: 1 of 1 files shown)


docs/user/README.md
```diff
@@ -16,9 +16,9 @@ The following commands are for Ubuntu Server. However, equivalent commands shoul
 1. Install [cURL](https://curl.se/) as the *root* user: `pkg install curl`
 
 ### Windows
-1. Install a supported version of [Visual Studio](https://visualstudio.microsoft.com/vs/older-downloads/) with Visual C++:
-   - [Visual Studio 2019](https://visualstudio.microsoft.com/vs/older-downloads/#visual-studio-2019-and-other-products) ([Supported](https://github.com/microsoft/snmalloc/blob/main/docs/BUILDING.md#building-on-windows))
-   - [Visual Studio 2022](https://visualstudio.microsoft.com/vs/) ([Supported](https://github.com/microsoft/snmalloc/blob/main/docs/BUILDING.md#building-on-windows))
+1. Install the Microsoft Visual C++ Prerequisites for Rust and Python 3:
+   - Microsoft Visual C++ Prerequisites for Rust: see [The rustup book](https://rust-lang.github.io/rustup/installation/windows-msvc.html).
```
CGodiksen, line 20: The linked guide mentions that Rust supports "Visual Studio 2013 and later" but since we have had issues using any other version than 2022, should we be explicit and say that they should install Visual Studio 2022? The walk-through in the guide uses 2022 so it should be fine for new installs but a user might think that an already installed version is sufficient.

docs/user/README.md
```diff
@@ -16,9 +16,9 @@ The following commands are for Ubuntu Server. However, equivalent commands shoul
...
+   - Python 3: install [Python 3](https://www.python.org/) and select _"Add python.exe to PATH"_.
```
CGodiksen, line 21: is any version of Python 3 fine or should we be more specific?

docs/user/README.md
skejserjensen: Generally I would prefer not to mention specific version number as we then have to continuously update them and since VS 2022 is used in the guide I decided not to mention a specific version in our documentation. What do you think @chrthomsen?

docs/user/README.md
skejserjensen: The documentation for `maturin` states that it works for [CPython3.5+](https://github.com/PyO3/maturin) and both [CPython 3.5 and CPython 3.6](https://devguide.python.org/versions/) is already end-of-life, so any version of CPython 3 that is still supported should be fine. So I think it is fine to not mention a specific CPython version as we support all versions that is not end-of-life. What do you think @chrthomsen?

docs/user/README.md
skejserjensen: Maybe we could simply add something like "the latest version" to Line 31 so it becomes "1. Install the latest version of the Microsoft Visual C++ Prerequisites for Rust and Python 3 "?

docs/user/README.md
skejserjensen: Maybe we could simply add something like "the latest version" to Line 31 so it becomes "1. Install the latest version of the Microsoft Visual C++ Prerequisites for Rust and Python 3 "?

docs/user/README.md
CGodiksen: Even though it was found to work with VS2019 I still think it makes sense to add "the latest versions" since it should still be preferable.

chrthomsen: I agree with adding "the latest versions"

docs/user/README.md
chrthomsen: Agreed again

chrthomsen (approved):

PR #120, opened 2023-07-26
Update dependencies

skejserjensen:
This PR primarily updates Apache Arrow DataFusion to [v28.0.0](https://crates.io/crates/datafusion/28.0.0) and fixes all known breaking changes. New versions of Apache Arrow DataFusion usually have a few breaking changes, so updating to each new release ensures that the number of changes required does not accumulate. `sqlx` was purposely not updated from [v0.6.3](https://crates.io/crates/sqlx/0.6.3) to [v0.7.1](https://crates.io/crates/sqlx/0.7.1) as it requires changes to `MetadataManager` and thus conflicts with `dev/common-metadata-manager`. Thus, `sqlx` should be updated in `dev/common-metadata-manager` or in a separate branch after `dev/common-metadata-manager` has been merged.


CGodiksen (approved):
I think it would be fine to force this through now. I can update the version of `sqlx` and make some changes to the manager `remote.rs` so it corresponds to the changes made here.

PR #121, opened 2023-08-03
Prepare for parallel StorageEngine

skejserjensen:
This PR performs some minor refactoring in preparation for making `StorageEngine` multithreaded, specifically it: 
- Makes `ModelTableMetadata` store `field_column_indices` so they only need to be computed once.
- It moves tracking of memory usage in `StorageEngine` to `MemoryPool` which is shared by the two managers. For example, this allows `UncompressedDataManager` to "allocate" memory for `UncompressedDataBuffer`s and for `CompressedDataManager` to  "free" the memory when it compresses the buffer.
- As both `Metric` and `MemoryPool` are fairly small types, they have been moved to a shared module named `types.rs`. While `CompressedDataBuffer` is also a fairly small type it has not been moved to use the same layout for everything related to management of uncompressed and compressed data.
- The `HashMap` used for caching tag hash in `MemoryManager` has been replaced with a thread safe `DashMap`. This should allow `MemoryManager` to efficiently be shared between multiple threads through an `Arc` without the overhead of a `RwLock` or a `Mutex`. Since this is a very minor change it should not conflict with `dev/common-metadata-manager`.


crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,291 @@
...
+            ),
+        }
+    }
+
+    /// Change the amount of memory available for storing
+    /// [`RecordBatches`](datafusion::arrow::record_batch::RecordBatch) containing multivariate time
+    /// series with metadata and
+    /// [`UncompressedDataBuffers`](crate::storage::uncompressed_data_buffer::UncompressedDataBuffer)
+    /// containing univariate time series without metadata by `size_in_bytes`.
+    pub(super) fn update_uncompressed_memory(&self, size_in_bytes: isize) {
```
CGodiksen, line 71: Maybe we should try to be consistent in the naming of methods that change a value like this. In another similar method (`adjust_uncompressed_remaining_memory_in_bytes`) we use `adjust` instead of `update`. I am not sure which is best, but we should probably use the same naming. 

If it is changed here, also change it in the other method.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,291 @@
...
+        *self.remaining_uncompressed_memory_in_bytes.read().unwrap()
+    }
+
+    /// Try to reserve `size_in_bytes` bytes of memory for storing
+    /// [`RecordBatches`](datafusion::arrow::record_batch::RecordBatch) containing multivariate time
+    /// series with metadata and
+    /// [`UncompressedDataBuffers`](crate::storage::uncompressed_data_buffer::UncompressedDataBuffer)
+    /// containing univariate time series without metadata. Returns [`true`] if the reservation
+    /// succeeds and [`false`] otherwise.
+    pub(super) fn try_reserve_uncompressed_memory(&self, size_in_bytes: usize) -> bool {
```
CGodiksen, line 92: Does it break Rust conventions to use `try_` with a function that return `bool` and not a `Result`? Personally I do not mind. 

If it is changed here, also change it in the other method.

crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -162,13 +162,16 @@ impl CompressedDataManager {
...
 
         // Update the remaining memory for compressed data and record the change.
-        self.compressed_remaining_memory_in_bytes -= segments_size as isize;
         self.used_compressed_memory_metric
             .append(segments_size as isize, true);
 
         // If the reserved memory limit is exceeded, save compressed data to disk.
-        if self.compressed_remaining_memory_in_bytes < 0 {
-            self.save_compressed_data_to_free_memory().await?;
+        if self
```
CGodiksen, line 169: Should this be negated since it returns [`true`] if the reservation succeeds which means we would want to free memory if the reservation did not succeed.

CGodiksen: Furthermore, if `try_reserve_compressed_memory` returns `false` i am not sure we ever retry to reserve the compressed memory after compressed data is saved. I think this was handled before by going to less than 0 before the check and then just saving until we were above 0 again.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+
+    /// Free `size_in_bytes` bytes of memory for storing
+    /// [`RecordBatches`](datafusion::arrow::record_batch::RecordBatch) containing multivariate time
+    /// series with metadata and
+    /// [`UncompressedDataBuffers`](crate::storage::uncompressed_data_buffer::UncompressedDataBuffer)
+    /// containing univariate time series without metadata. Returns [`true`] if the reservation
+    /// succeeds and [`false`] otherwise.
+    pub(super) fn free_uncompressed_memory(&self, size_in_bytes: usize) {
```
CGodiksen, line 113: It can be kept for convenience but since this method is very similar to `update_uncompressed_memory`, an argument could be made for removing this and only using `update_uncompressed_memory`.  

If this is removed, also remove `free_compressed_memory`.

CGodiksen (changes requested):

crates/modelardb_server/src/remote.rs
```diff
@@ -802,8 +802,8 @@ impl FlightService for FlightServiceHandler {
                 "compressed_reserved_memory_in_bytes",
             ];
             let values = [
-                *configuration_manager.uncompressed_reserved_memory_in_bytes() as u64,
-                *configuration_manager.compressed_reserved_memory_in_bytes() as u64,
+                configuration_manager.uncompressed_reserved_memory_in_bytes() as u64 ,
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -37,6 +37,8 @@ use crate::storage::compressed_data_buffer::CompressedDataBuffer;
 use crate::storage::data_transfer::DataTransfer;
 use crate::storage::{Metric, StorageEngine, COMPRESSED_DATA_FOLDER};
 
+use super::types::MemoryPool;
```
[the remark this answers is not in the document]
skejserjensen: Fixed, was an accident due to the import being automatically added by `rust-analyzer`.

crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -269,13 +272,16 @@ impl CompressedDataManager {
         Ok(table_relevant_apache_parquet_files)
     }
 
-    /// Save [`CompressedDataBuffers`](CompressedDataBuffer) to disk until the reserved memory limit
-    /// is no longer exceeded. If all of the data is saved successfully, return [`Ok`], otherwise
+    /// Save [`CompressedDataBuffers`](CompressedDataBuffer) to disk until at least `size_in_bytes`
+    /// of memory is available . If all of the data is saved successfully, return [`Ok`], otherwise
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -31,8 +32,6 @@ use std::fs::File;
 use std::io::{Error as IOError, Write};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
-use std::{fmt, mem};
 
 use crate::configuration::ConfigurationManager;
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -51,9 +52,8 @@ pub(super) struct UncompressedDataManager {
     /// FIFO queue of full [`UncompressedDataBuffers`](UncompressedDataBuffer) that are ready for
     /// compression.
     finished_uncompressed_data_buffers: VecDeque<Box<dyn UncompressedDataBuffer>>,
-    /// How many bytes of memory that are left for storing
-    /// [`UncompressedDataBuffers`](UncompressedDataBuffer).
-    uncompressed_remaining_memory_in_bytes: usize,
+    /// Track how much memory are left for storing uncompressed and compressed data.
```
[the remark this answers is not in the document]
skejserjensen: As far as I am aware uncountable nouns is always singular, so I agree that "memory" should be "is".

crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -51,10 +53,8 @@ pub(super) struct CompressedDataManager {
     /// FIFO queue of table names and column indices referring to [`CompressedDataBuffer`] that can
     /// be saved to persistent storage.
     compressed_queue: VecDeque<(String, u16)>,
-    /// How many bytes of memory that are left for storing compressed segments. A signed integer is
-    /// used since compressed data is inserted and then the remaining bytes are checked. This means
-    /// that the remaining bytes can briefly be negative until compressed data is saved to disk.
-    compressed_remaining_memory_in_bytes: isize,
+    /// Track how much memory are left for storing uncompressed and compressed data.
```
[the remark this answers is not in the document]
skejserjensen: As far as I am aware uncountable nouns is always singular, so I agree that "memory" should be "is".

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,291 @@
...
+pub(super) struct MemoryPool {
+    /// How many bytes of memory that are left for storing
+    /// [`RecordBatches`](datafusion::arrow::record_batch::RecordBatch) containing multivariate time
+    /// series with metadata and
+    /// [`UncompressedDataBuffers`](crate::storage::uncompressed_data_buffer::UncompressedDataBuffer)
+    /// containing univariate time series without metadata.
+    remaining_uncompressed_memory_in_bytes: RwLock<isize>,
+    /// How many bytes of memory that are left for storing
+    /// [`CompressedDataBuffers`](crate::storage::compressed_data_buffer::CompressedDataBuffer). A
+    /// signed integer is used since compressed data is inserted and then the remaining bytes are
```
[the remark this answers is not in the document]
skejserjensen: I have moved it to the doc comment for `MemoryPool`.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,291 @@
...
+    /// checked. This means that the remaining bytes can briefly be negative until compressed data
+    /// is saved to disk and if the amount of memory for uncompressed data is reduced.
```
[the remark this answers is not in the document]
skejserjensen: The text has been revised.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,291 @@
...
+    remaining_compressed_memory_in_bytes: RwLock<isize>,
+}
+
+/// The pool of memory the
+/// [`UncompressedDataManager`](crate::storage::uncompressed_data_manager::UncompressedDataManager)
+/// can use for uncompressed data and the
+/// [`CompressedDataManager`](crate::storage::CompressedDataManager) can use for compressed data.
+impl MemoryPool {
+    /// Create a new [`MemoryPool`] with at most [`u64::MAX`] bytes of memory for uncompressed data
```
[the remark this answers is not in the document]
skejserjensen: Fixed, it was a leftover from a previous design.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,291 @@
...
+                uncompressed_memory_in_bytes.try_into().unwrap(),
+            ),
+            remaining_compressed_memory_in_bytes: RwLock::new(
+                compressed_memory_in_bytes.try_into().unwrap(),
+            ),
+        }
+    }
+
+    /// Change the amount of memory available for storing
+    /// [`RecordBatches`](datafusion::arrow::record_batch::RecordBatch) containing multivariate time
```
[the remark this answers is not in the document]
skejserjensen: Changed the explanation to "uncompressed data" and "compressed data".

crates/modelardb_server/src/storage/types.rs
skejserjensen: I changed `update_` to `adjust_` in `MemoryPool` as it required the fewest changes and I do not have strong opinion on `adjust` vs `update`.

crates/modelardb_server/src/storage/types.rs
skejserjensen: While `try_` generally indicate that a method can fail and most return a `Result`, I am not aware of any convention that it should return `Result`. I used `bool` instead of `Result` as it made the API nicer to use in my opinion, but we can change to an error type that indicates that memory could not be reserved and then use `if ....is_err()` if you think it is beneficial?

crates/modelardb_server/src/storage/compressed_data_manager.rs
skejserjensen: Fixed by changing it to `while ! ...`.

crates/modelardb_server/src/storage/types.rs
skejserjensen: I also contemplated removing the `free_` methods and using the `update_`/`adjust_` methods for both freeing and adjusting memory, but I kept the `free_` methods as they can take an `usize` as input (freeing a negative amount of memory is not a thing) and thus prevents accidental freeing a negative amount of memory and the amount of code duplication is very small. However, the trade-off should definitely be reevaluated if the methods are extended.

crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -473,7 +475,7 @@ impl UncompressedDataManager {
             {
                 // Add the size of the in-memory data buffer back to the remaining reserved bytes.
                 let freed_memory = UncompressedInMemoryDataBuffer::memory_size();
-                self.uncompressed_remaining_memory_in_bytes += freed_memory;
+                self.memory_pool.free_uncompressed_memory(freed_memory);
                 self.used_uncompressed_memory_metric
```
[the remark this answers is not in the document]
skejserjensen: I have added a low priority task.

crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -269,13 +271,16 @@ impl CompressedDataManager {
         Ok(table_relevant_apache_parquet_files)
     }
 
-    /// Save [`CompressedDataBuffers`](CompressedDataBuffer) to disk until the reserved memory limit
-    /// is no longer exceeded. If all of the data is saved successfully, return [`Ok`], otherwise
-    /// return [`IOError`].
-    async fn save_compressed_data_to_free_memory(&mut self) -> Result<(), IOError> {
+    /// Save [`CompressedDataBuffers`](CompressedDataBuffer) to disk until at least `size_in_bytes`
+    /// bytes of memory is available . If all of the data is saved successfully, return [`Ok`],
```
CGodiksen, line 275: Very very minor but it seems there is an extra space after `available`.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,478 @@
...
+        );
+
+        assert_eq!(
+            memory_pool.remaining_uncompressed_memory_in_bytes(),
+            -(common_test::UNCOMPRESSED_RESERVED_MEMORY_IN_BYTES as isize)
+        );
+    }
+
+    #[test]
+    fn test_try_reserve_uncompressed_memory_succeed() {
```
CGodiksen, line 275: It seems odd to have `succeed` and `fail` in test names. It might be better to name the tests so they include why it should succeed or fail. Maybe rename the tests to something like `test_try_reserve_available_uncompressed_memory` and `test_try_reserve_unavailable_uncompressed_memory` or something similar. 

If the change is made here, also change it for the corresponding tests for compressed.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -0,0 +1,478 @@
...
+        let memory_pool = create_memory_pool();
+        assert_eq!(
+            memory_pool.remaining_uncompressed_memory_in_bytes(),
+            common_test::UNCOMPRESSED_RESERVED_MEMORY_IN_BYTES as isize
+        );
+
+        memory_pool
```
CGodiksen, line 282: To test that the function both makes the correct change and returns the correct boolean value, maybe add an `assert!` around all the calls to `memory_pool.try_reserve_uncompressed_memory` and `memory_pool.try_reserve_compressed_memory` in the tests.

crates/modelardb_server/src/storage/types.rs
CGodiksen: As it does not go against any conventions I also think the current API is fine.

crates/modelardb_server/src/storage/compressed_data_manager.rs
skejserjensen: Fixed.

crates/modelardb_server/src/storage/types.rs
skejserjensen: Fixed.

crates/modelardb_server/src/storage/types.rs
skejserjensen: Good idea, fixed and added the [`must_use` attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute).

CGodiksen (approved):

chrthomsen (approved):

PR #123, opened 2023-08-08
Add \d, \dt, and SQL to user README Execute SQL

skejserjensen:
This PR extends docs/userREADME by adding examples of \d, \dt, and SQL to the section Execute SQL as \d and \dt was not documented.

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -120,7 +120,7 @@ modelardb
 
 Be aware that the REPL currently does not support splitting SQL statements over multiple lines and that the SQL
 statements do not need to end with a `;`. In addition to SQL statements, the REPL also supports listing all tables
-using `\dt` and printing the schema of a table using`\d table_name`:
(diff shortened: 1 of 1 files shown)


docs/user/README.md
```diff
@@ -118,6 +118,31 @@ local instance of `modelardbd` through a REPL, simply run `modelardb`:
 modelardb
 ```
 
+Be aware that the REPL currently does not support splitting SQL statements over multiple lines and that the SQL
+statements do not need to end with a `;`. In addition to SQL statements, the REPL also supports listing all tables
+using `\dt` and printing the schema of a table using`\d table_name`:
```
CGodiksen, line 123: Very minor but it seems that a space is missing after `using`.

CGodiksen (changes requested):

docs/user/README.md
skejserjensen: Fixed.

CGodiksen (approved):

PR #127, opened 2023-09-01
Fixed validation of Azure Blob Storage object stores

CGodiksen:
This PR fixes a bug that made it possible to start the server with an Azure Blob Storage object store with valid connection information but an invalid container name. Object stores are now only seen as valid if the connection information is correct and the container/bucket exists. 

Since it is no longer necessary to pass the remote data folder type, the struct has been removed as well as the corresponding utility function. Furthermore, since the interface is simpler, the server now also validates the object store it receives from the manager if started in `MultiNode` mode.

### DIFF
--- a/crates/modelardb_common/src/arguments.rs
+++ b/crates/modelardb_common/src/arguments.rs
@@ -129,28 +129,9 @@ impl FromStr for RemoteDataFolderType {
     }
 }
 
-/// Extract the remote data folder type from the arguments and validate that the remote data folder can be
-/// accessed. If the remote data folder cannot be accessed, return the error that occurred as a [`String`].
-pub async fn validate_remote_data_folder_from_argument(
-    argument: &str,
-    remote_data_folder: &Arc<dyn ObjectStore>,
-) -> Result<(), String> {
-    if let Some(split_argument) = argument.split_once("://") {
-        let object_store_type = split_argument.0;
-        let remote_data_folder_type = RemoteDataFolderType::from_str(object_store_type)?;
-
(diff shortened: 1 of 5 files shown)


crates/modelardb_common/src/arguments.rs
```diff
@@ -106,63 +106,26 @@ pub fn argument_to_connection_info(argument: &str) -> Result<Vec<u8>, String> {
...
-    }
-}
-
 /// Validate that the remote data folder can be accessed. If the remote data folder cannot be
 /// accessed, return the error that occurred as a [`String`].
 pub async fn validate_remote_data_folder(
-    remote_data_folder_type: RemoteDataFolderType,
     remote_data_folder: &Arc<dyn ObjectStore>,
 ) -> Result<(), String> {
+    let invalid_path = Uuid::new_v4().to_string();
```
skejserjensen, line 114: Maybe add a comment why a UUID is used as `invalid_path` instead of a pre-generated random path so it is clear if we need to change the code at a later date, I assume it is to minimize the change of using a filename that exists.

skejserjensen (changes requested):
I have also extended the system test task to add a system test that checks if the strings "BlobNotFound" and "NoSuchKey" are changed when the environment for system tests are added.

crates/modelardb_common/src/arguments.rs
CGodiksen: Fixed.

skejserjensen (approved):

chrthomsen (approved):

PR #128, opened 2023-09-05
New integration tests for existing actions

CGodiksen:
This PR adds a few integration tests to the actions `CollectMetrics` and `UpdateConfiguration`/`GetConfiguration`. These actions were previously not tested properly since the actions return the response in raw bytes and we were not able to find a method to read these bytes into a record batch. 

This PR adds a method to do this and adds a few tests that use this method to finally test the actions.

### DIFF
--- a/crates/modelardb_server/tests/integration_test.rs
+++ b/crates/modelardb_server/tests/integration_test.rs
@@ -32,7 +32,7 @@ use std::time::Duration;
 use arrow_flight::flight_service_client::FlightServiceClient;
 use arrow_flight::{utils, Action, Criteria, FlightData, FlightDescriptor, PutResult, Ticket};
 use bytes::{Buf, Bytes};
-use datafusion::arrow::array::{Array, ListArray, StringArray};
+use datafusion::arrow::array::{Array, ListArray, StringArray, UInt64Array};
 use datafusion::arrow::compute;
 use datafusion::arrow::datatypes::{DataType, Field, Schema, TimeUnit::Millisecond};
 use datafusion::arrow::ipc::convert;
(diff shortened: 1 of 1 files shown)


crates/modelardb_server/tests/integration_test.rs
```diff
@@ -586,22 +613,38 @@ fn test_can_list_actions() {
...
+    let metrics_array = modelardb_common::array!(metrics, 0, StringArray);
+
+    assert_eq!(metrics_array.value(0), "used_uncompressed_memory");
+    assert_eq!(metrics_array.value(1), "used_compressed_memory");
+    assert_eq!(metrics_array.value(2), "ingested_data_points");
+    assert_eq!(metrics_array.value(3), "used_disk_space");
+
+    // Check that the metrics are populated when ingesting and flushing.
+    let timestamps_array = modelardb_common::array!(metrics, 1, ListArray);
+    assert_eq!(timestamps_array.value(0).len(), 2);
```
skejserjensen, line 644: Maybe match the specific lists when possible and document why only their length can be checked for the rest.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -763,6 +806,32 @@ fn test_optimized_query_results_equals_non_optimized_query_results() {
     assert_eq!(optimized_query, non_optimized_query);
 }
 
+#[test]
+fn test_can_update_uncompressed_reserved_memory_in_bytes() {
```
skejserjensen, line 810: Maybe refactor the code in these two tests into a shared function as the two tests only seem to differ by a string and an index.

skejserjensen (changes requested):

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -425,6 +426,32 @@ impl TestContext {
         self.runtime
             .block_on(async { self.client.do_action(Request::new(action)).await })
     }
+
+    /// Retrieve the response of the action with the type `action_type` and convert it into an
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

crates/modelardb_server/tests/integration_test.rs
CGodiksen: Fixed.

crates/modelardb_server/tests/integration_test.rs
CGodiksen: Good idea, fixed.

skejserjensen (approved):

PR #130, opened 2023-09-11
Fixed broken documentation link and made other links local

CGodiksen:
This PR fixes a link in the user documentation that linked to a file in the repository that did not exist. It also makes some links to files in the repository use a local path instead of using the `Github` path. This makes it a bit more future proof since the links will not break if the repository host is moved and it also ensures that we link to the files in the current branch and not always the master branch.

### DIFF
--- a/crates/modelardb_compression_python/README.md
+++ b/crates/modelardb_compression_python/README.md
@@ -8,7 +8,7 @@ and
 [jhoekx/python-rust-arrow-interop-example](https://github.com/jhoekx/python-rust-arrow-interop-example/blob/master/src/lib.rs).
 
 ## Installation
-1. Install the [dependencies](https://github.com/ModelarData/ModelarDB-RS#installation) required to build ModelarDB's compression library.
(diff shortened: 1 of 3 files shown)


skejserjensen (changes requested):
The links to Minio's web-interface also seems to be broken as they have switched to selecting a [random port for the web-interface](https://min.io/docs/minio/linux/administration/minio-console.html#static-vs-dynamic-port-assignment) at startup. Maybe remove the links and tell users to look at the output of `minio server folder` or tell users to include `--console-address ":9001"` as specified in Minio's documentation linked above.

CGodiksen:
> 

I have removed the links to the web interface in the main documentation since we make no assumptions about the users usage of MinIO in that part of the documentation. In the Docker section I have kept the link since we assume that they are using our Docker compose file which starts the MinIO server with `--console-address ":9001"` already.

skejserjensen (approved):

chrthomsen (approved):

PR #137, opened 2023-10-22
Fix minor consistency errors in StorageEngine

skejserjensen:
This PR fixes minor consistency errors in `StorageError`. Thus, the PR purposely converts the two functions in `storage/mod.rs` into methods on `StorageEngine` as it required the smallest amount of changes to make the implementation of sub-routines that do not have a `self` parameter consistent in `StorageEngine`.

### DIFF
--- a/crates/modelardb_server/src/storage/mod.rs
+++ b/crates/modelardb_server/src/storage/mod.rs
@@ -83,7 +83,8 @@ pub(super) const COMPRESSED_DATA_FOLDER: &str = "compressed";
 pub(super) const QUERY_DATA_FOLDER_SCHEME_WITH_HOST: &str = "query://query";
 
 /// A static UUID for use in tests. It is not in a test utilities module so it can be used both
-/// inside the if cfg(test) block of [`create_time_and_value_range_file_name`] and in test modules.
+/// inside the if cfg(test) block of [`StorageEngine::create_time_and_value_range_file_name`] and in
+/// test modules.
 const TEST_UUID: Uuid = uuid!("44c57d06-333c-4935-8ae3-ed7bc53a08c4");
 
 /// The expected [first four bytes of any Apache Parquet file].
@@ -383,7 +384,7 @@ impl StorageEngine {
         // Retrieve object_metas that represent the relevant files for table_name and column_index.
         let relevant_apache_parquet_files = self
             .compressed_data_manager
-            .get_saved_compressed_files(
+            .compressed_files(
                 table_name,
                 column_index,
                 start_time,
--- a/crates/modelardb_server/src/storage/mod.rs
+++ b/crates/modelardb_server/src/storage/mod.rs
@@ -83,8 +83,8 @@ pub(super) const COMPRESSED_DATA_FOLDER: &str = "compressed";
 pub(super) const QUERY_DATA_FOLDER_SCHEME_WITH_HOST: &str = "query://query";
 
 /// A static UUID for use in tests. It is not in a test utilities module so it can be used both
-/// inside the if cfg(test) block of [`StorageEngine::create_time_and_value_range_file_name`] and in
-/// test modules.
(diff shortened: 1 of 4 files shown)


crates/modelardb_server/src/storage/mod.rs
```diff
@@ -74,17 +74,18 @@ use crate::storage::uncompressed_data_manager::UncompressedDataManager;
...
+pub(super) const COMPRESSED_DATA_FOLDER: &str = "compressed";
 
 /// The scheme with host at which the query data folder is stored.
-pub const QUERY_DATA_FOLDER_SCHEME_WITH_HOST: &str = "query://query";
+pub(super) const QUERY_DATA_FOLDER_SCHEME_WITH_HOST: &str = "query://query";
 
 /// A static UUID for use in tests. It is not in a test utilities module so it can be used both
-/// inside the if cfg(test) block of [`create_time_and_value_range_file_name`] and in test modules.
-pub const TEST_UUID: Uuid = uuid!("44c57d06-333c-4935-8ae3-ed7bc53a08c4");
+/// inside the if cfg(test) block of [`StorageEngine::create_time_and_value_range_file_name`] and in
```
CGodiksen, line 86: Minor but maybe fix the link to remove `StorageEngine::` for consistency.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -624,8 +628,62 @@ impl StorageEngine {
             .map_err(|error| ParquetError::General(error.to_string()))
     }
 
+    /// Create a file name that includes the start timestamp of the first segment in `batch`, the
```
CGodiksen, line 631: I am not opposed to making this change but it seems that we once agreed that we should use "free" functions instead of associated functions like here, whenever possible, according to [this](https://dev.azure.com/ModelarData/ModelarDB%20Implementation/_workitems/edit/180) work item. 

Maybe either delete the work item or change it to make it clear we are doing the opposite.

CGodiksen (changes requested):

crates/modelardb_server/src/storage/mod.rs
CGodiksen, line 86: Maybe change to `StorageEngine::create_time_and_value_range_file_name()` to make it clear that it is a function. This would also be consistent with the doc comment for this file.

skejserjensen: `StorageEngine::` is needed as `create_time_and_value_range_file_name()` is currently a method in `StorageEngine` that simply does not take `self` as as an argument.

crates/modelardb_server/src/storage/mod.rs
skejserjensen: I tried moving methods out of `StorageEngine` but it created a larger PR then moving the two functions into `StorageEngine`. Despite going against what we previously discussed, I tried to refactor as little code as possible to make it easier to review the collection of small changes made in this PR. This way `StorageEngine` is now, as far as I can tell,` consistent and we can make a PR later that only moves the methods that does not take `self` as a parameter into functions and thus should be simple to review since it only changes code without changing it.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: I removed the old comment and added this one that just says to add `()`. I realized `StorageEngine` was needed so I deleted it.

CGodiksen (changes requested):
The minor comment asking to add `()` was not fixed.

crates/modelardb_server/src/storage/mod.rs
skejserjensen: Fixed.

CGodiksen (approved):

chrthomsen (approved):

PR #145, opened 2023-11-11
Configurable transfer batch size

CGodiksen:
To make it easier to setup ModelarDB for any use-case this PR moves the transfer batch size in bytes from a constant to the configuration and makes it possible to update it while the server is running. Furthermore, the configurable settings can now also be set through environment variables before the server is started, while still using the same defaults if the environment variables are not set. 

A description of the available environment variables has also been added to the user documentation.


crates/modelardb_manager/src/main.rs
```diff
@@ -42,13 +42,8 @@ use crate::metadata::MetadataManager;
...
 /// The port of the Apache Arrow Flight Server. If the environment variable is not set, 8888 is used.
-pub static PORT: Lazy<u16> = Lazy::new(|| match env::var("MODELARDBM_PORT") {
-    Ok(port) => port
-        .parse()
-        .map_err(|_| "MODELARDBM_PORT must be between 1 and 65535.")
-        .unwrap(),
-    Err(_) => 8888,
-});
+pub static PORT: Lazy<u16> =
+    Lazy::new(|| env::var("MODELARDBM_PORT").map_or(8888, |value| value.parse().unwrap()));
```
skejserjensen, line 46: Maybe use a port adjacent to `modelardbd`s default port to make it simpler to remember the ports used by the different components of ModelarDB. Also the commonly used Jupyter notebooks [use port 8888 by default](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) so maybe use a different pot to allow both to be run together.

crates/modelardb_server/src/storage/data_transfer.rs
```diff
@@ -153,19 +150,35 @@ impl DataTransfer {
...
     }
 
-    /// Transfer all compressed files currently in the data folder to the remote object store.
-    /// Return [`Ok`] if all files were transferred successfully, otherwise [`ParquetError`]. Note
-    /// that if the function fails, some of the compressed files may still have been transferred.
-    /// Since the data is transferred separately for each table, the function can be called again if
-    /// it failed.
-    pub(crate) async fn flush(&self) -> Result<(), ParquetError> {
+    /// Set the transfer batch size to `new_value`. For each table that compressed data is saved
+    /// for, check if the amount of data exceeds `new_value` and transfer if so. If the value
```
skejserjensen, line 154: Maybe "... exceeds `new_value` and transfer data if it is." Also maybe document how much data is transferred, is it all of the data or just enough to get below `new_value`.

crates/modelardb_server/src/storage/data_transfer.rs
```diff
@@ -489,6 +502,60 @@ mod tests {
...
+            .set_transfer_batch_size_in_bytes(COMPRESSED_FILE_SIZE * 10)
+            .await
+            .unwrap();
+
+        assert_eq!(
+            data_transfer.transfer_batch_size_in_bytes,
+            COMPRESSED_FILE_SIZE * 10
+        );
+
+        // Data should not be transferred.
```
skejserjensen, line 526: Maybe "Data should not have been transferred."

crates/modelardb_server/src/storage/data_transfer.rs
```diff
@@ -500,7 +567,10 @@ mod tests {
         let (target_dir, data_transfer) =
             create_data_transfer_component(metadata_manager, temp_dir.path()).await;
 
-        data_transfer.flush().await.unwrap();
+        data_transfer
```
skejserjensen, line 570: Unclear why `data_transfer.flush()` was replaced with`transfer_larger_than_threshold(0)` maybe add a comment or a `flush()` method to `data_transfer()` that is simply executes `transfer_larger_than_threshold(0)` and use that everywhere for consistency with the other storage components as they implements `flush()`.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -466,6 +476,23 @@ impl StorageEngine {
...
+        new_value: usize,
+    ) -> Result<(), ParquetError> {
+        if let Some(ref mut data_transfer) =
+            *self.compressed_data_manager.data_transfer.write().await
+        {
+            data_transfer
+                .set_transfer_batch_size_in_bytes(new_value)
+                .await
+        } else {
+            Ok(())
```
skejserjensen, line 492: Maybe return an error stating that no data transfer component has been created instead of failing silently. If a user tries to changes the data transfer batch size on a node they probably think that a data transfer component exists.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -680,15 +680,26 @@ fn test_can_collect_metrics() {
...
+            uncompressed_buffer_size,
+            uncompressed_buffer_size * 2,
+            uncompressed_buffer_size * 3,
+            uncompressed_buffer_size * 4,
+            uncompressed_buffer_size * 5,
+            uncompressed_buffer_size * 4,
+            uncompressed_buffer_size * 3,
+            uncompressed_buffer_size * 2,
+            uncompressed_buffer_size,
+            0
```
skejserjensen, line 701: Maybe `uncompressed_buffer_size * 0` to try and make it more clear why it is zero.

docs/user/README.md
```diff
@@ -252,6 +252,18 @@ By using Telegraf, data points can be efficiently streamed into `modelardbd` fro
...
+|--------------------------------------------------|-------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| MODELARDBD_PORT                                  | 9999        | The port of the Apache Arrow Flight Server.                                                                                                                                                                                                                                                                                                                |
```
skejserjensen, line 261: If Apache Arrow Flight endpoint is in #144 maybe replace "Server" with "Endpoint" vice versa.

docs/user/README.md
```diff
@@ -252,6 +252,18 @@ By using Telegraf, data points can be efficiently streamed into `modelardbd` fro
...
+| MODELARDBD_UNCOMPRESSED_RESERVED_MEMORY_IN_BYTES | 512 MB      | Amount of memory to reserve for storing uncompressed data buffers.                                                                                                                                                                                                                                                                                         |
```
skejserjensen, line 262: Maybe prefix with "The" to match the other settings.

docs/user/README.md
```diff
@@ -252,6 +252,18 @@ By using Telegraf, data points can be efficiently streamed into `modelardbd` fro
...
+| MODELARDBD_COMPRESSED_RESERVED_MEMORY_IN_BYTES   | 512 MB      | Amount of memory to reserve for storing compressed data buffers.                                                                                                                                                                                                                                                                                           |
```
skejserjensen, line 263: Maybe prefix with "The" to match the other settings.

docs/user/README.md
```diff
@@ -252,6 +252,18 @@ By using Telegraf, data points can be efficiently streamed into `modelardbd` fro
...
+| MODELARDBD_TRANSFER_BATCH_SIZE_IN_BYTES          | 64 MB       | The number of bytes that are required before transferring a batch of  data to the remote object store.                                                                                                                                                                                                                                                     |
```
skejserjensen, line 264: Maybe write something like "The amount of data that must be collected before transferring...."

docs/user/README.md
```diff
@@ -252,6 +252,18 @@ By using Telegraf, data points can be efficiently streamed into `modelardbd` fro
...
+| MODELARDBD_UNCOMPRESSED_DATA_BUFFER_CAPACITY     | 65536       | The capacity of each uncompressed data buffer as the number of elements in the buffer where each element is a timestamp and a value. Note that the resulting size of the buffer has to be a multiple of 64 bytes to avoid the actual capacity being larger than the requested due to internal alignment when allocating memory for the two array builders. |
```
skejserjensen, line 265: Maybe remove "...when allocating memory for the two array builders." as it seem to be too implementation specific for the user documentation.

skejserjensen (changes requested):

crates/modelardb_server/src/storage/data_transfer.rs
CGodiksen: I think it is overkill to provide such methods in components such as the data transfer component since it quite deep in the storage engine architecture and not accessed directly outside of it.

crates/modelardb_manager/src/main.rs
CGodiksen: Fixed.

crates/modelardb_server/src/storage/data_transfer.rs
CGodiksen: Fixed.

crates/modelardb_server/src/storage/data_transfer.rs
CGodiksen: Fixed.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: Fixed.

crates/modelardb_server/tests/integration_test.rs
CGodiksen: Fixed.

docs/user/README.md
CGodiksen: I am not sure what you mean. The server has endpoints but the port is related to the server. I have left it as it was.

docs/user/README.md
CGodiksen: Fixed.

docs/user/README.md
CGodiksen: Fixed.

docs/user/README.md
CGodiksen: Fixed.

docs/user/README.md
CGodiksen: Fixed.

docs/user/README.md
skejserjensen: In #144 I suggested calling it an Apache Arrow Flight endpoint (like `AWS_ENDPOINT`) to differentiate it from `modelardbd` which we have been calling the server in the [documentation](https://github.com/ModelarData/ModelarDB-RS/blob/main/docs/user/README.md?plain=1#L29) and in the [source code](https://github.com/ModelarData/ModelarDB-RS/blob/main/crates/modelardb_common/src/types.rs#L105). If you think it is clear enough if we use server for both `modelardbd` and Apache Arrow Flight that is also fine for me. Thus, I have kept this thread unresolved but approved the PR. Feel free to resolve the thread after making a decision on what term to use for Apache Arrow Flight.

skejserjensen (approved):

crates/modelardb_server/tests/integration_test.rs
skejserjensen: Sorry, I was not aware that Clippy warns about multiplication by zero.

skejserjensen (approved):

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -78,10 +80,17 @@ pub(super) const QUERY_DATA_FOLDER_SCHEME_WITH_HOST: &str = "query://query";
...
 /// size of the buffer has to be a multiple of 64 bytes to avoid the actual capacity being larger
 /// than the requested due to internal alignment when allocating memory for the two array builders.
-const UNCOMPRESSED_DATA_BUFFER_CAPACITY: usize = 64 * 1024;
+pub static UNCOMPRESSED_DATA_BUFFER_CAPACITY: Lazy<usize> = Lazy::new(|| {
+    env::var("MODELARDBD_UNCOMPRESSED_DATA_BUFFER_CAPACITY").map_or(64 * 1024, |value| {
+        let parsed = value.parse::<usize>().unwrap();
 
-/// The number of bytes that are required before transferring a batch of data to the remote object store.
-const TRANSFER_BATCH_SIZE_IN_BYTES: usize = 64 * 1024 * 1024; // 64 MiB;
+        if parsed >= 64 && parsed % 64 == 0 {
```
chrthomsen, line 87: is 64 too small to be practically useful? Should there a be minimum value that we will use instead?

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -78,10 +80,17 @@ pub(super) const QUERY_DATA_FOLDER_SCHEME_WITH_HOST: &str = "query://query";
...
+            parsed
+        } else {
+            panic!("MODELARDBD_UNCOMPRESSED_DATA_BUFFER_CAPACITY must be a multiple of 64.")
```
chrthomsen, line 90: should it panic or fall back to the closest multiple of 64?

docs/user/README.md
```diff
@@ -252,6 +252,18 @@ By using Telegraf, data points can be efficiently streamed into `modelardbd` fro
...
+| MODELARDBD_COMPRESSED_RESERVED_MEMORY_IN_BYTES   | 512 MB      | The amount of memory to reserve for storing compressed data buffers.                                                                                                                                                                                                                                     |
+| MODELARDBD_TRANSFER_BATCH_SIZE_IN_BYTES          | 64 MB       | The amount of data that must be collected before transferring a batch of data to the remote object store.                                                                                                                                                                                                |
+| MODELARDBD_UNCOMPRESSED_DATA_BUFFER_CAPACITY     | 65536       | The capacity of each uncompressed data buffer as the number of elements in the buffer where each element is a timestamp and a value. Note that the resulting size of the buffer has to be a multiple of 64 bytes to avoid the actual capacity being larger than the requested due to internal alignment. |
```
chrthomsen, line 265: use KB?

chrthomsen (approved):

crates/modelardb_server/src/storage/mod.rs
CGodiksen: I think it is fine to just have the limit be a functional limit instead of a practical one for now. We probably have to look into the issue of more realistic settings in general later.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: This panic would happen when they start the server so not much would be lost so I think it is fine to just panic.

docs/user/README.md
CGodiksen: The reason we use a full number instead of KB is to avoid this capacity being confused with the memory usage of the buffer.

PR #151, opened 2023-12-20
Added time-based transfer of data

CGodiksen:
This PR adds support for time-based data transfer in the data transfer component. It is now possible to set a setting in the configuration that controls how many seconds there should be between each periodic data transfer. Furthermore, it is now possible to disable both size-based and time-based data transfer by updating the corresponding setting in the configuration with an empty value.


crates/modelardb_server/src/storage/data_transfer.rs
```diff
@@ -160,14 +170,67 @@ impl DataTransfer {
...
+                            .await
+                            .expect("Periodic data transfer failed.");
+                    }
+                });
+
+            // Start a tokio task that checks if the scheduler has pending tasks and runs them if so.
+            let join_handle = tokio::spawn(async move {
+                loop {
+                    scheduler.run_pending().await;
+                    tokio::time::sleep(Duration::from_millis(1000)).await;
```
skejserjensen, line 224: Could we remove this extra sleep so when the data is transferred only depends on transfer time in seconds instead of two timers and also remove `clockwerk` by simply creating a [`tokio::task`](https://docs.rs/tokio/latest/tokio/task/index.html) that sleeps for `transfer_time_in_seconds` and calls `transfer_larger_than_threshold(0)`? Since that would also create a [`JoinHandle`](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html) it should still be possible to use `task.abort()` when updating transfer time in seconds.

crates/modelardb_server/src/storage/data_transfer.rs
```diff
@@ -457,6 +520,34 @@ mod tests {
...
+
+        let (_, mut data_transfer) =
+            create_data_transfer_component(metadata_manager, temp_dir.path()).await;
+
+        data_transfer
+            .set_transfer_batch_size_in_bytes(None)
+            .await
+            .unwrap();
+
+        assert_eq!(data_transfer.transfer_batch_size_in_bytes, None);
```
skejserjensen, line 539: Maybe check the value is not `None` the call to `set_transfer_batch_size_in_bytes(None)`.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -228,13 +228,22 @@ impl StorageEngine {
...
             compressed_data_manager,
             memory_pool,
             join_handles,
             channels,
-        })
+        };
+
+        // Start the task that transfers data periodically if a data transfer component exists and
+        // time-based data transfer is enabled. Errors are ignored since an error is only returned
+        // if a data transfer component does not exist.
```
skejserjensen, line 241: Maybe change "if a data transfer component does not exist." to something like "if the data transfer component has not been enabled." as it always exists.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -228,13 +228,22 @@ impl StorageEngine {
...
+        let _ = storage_engine
```
skejserjensen, line 242: Maybe check if the data transfer component is enabled before calling the method and return all errors returned by `set_transfer_time_in_seconds(...)` as ignoring all errors may create issues that are hard to debug in the future, e.g., if we start returning additional errors which are then also ignored which we will have forgotten about at that time.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -986,51 +990,71 @@ fn update_and_retrieve_configuration_values(setting: &str) -> UInt64Array {
...
         );
     }
 }
 
 #[test]
 fn test_cannot_update_setting_with_invalid_value() {
+    update_configuration_and_assert_error(
+        "compressed_reserved_memory_in_bytes",
+        "-1",
+        "New value for compressed_reserved_memory_in_bytes is not valid: invalid digit found in string",
```
skejserjensen, line 1050: Instead of "invalid digit found in string" maybe write the supported range.

skejserjensen (changes requested):

crates/modelardb_server/src/configuration.rs
```diff
@@ -241,7 +269,28 @@ mod tests {
...
-            1024
+            Some(1024)
+        );
+    }
+
+    #[tokio::test]
+    async fn test_set_transfer_time_in_seconds() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let (storage_engine, configuration_manager) = create_components(temp_dir.path()).await;
+
```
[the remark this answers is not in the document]
CGodiksen: Fixed in this PR.

crates/modelardb_server/src/remote.rs
```diff
@@ -598,29 +603,50 @@ impl FlightService for FlightServiceHandler {
         } else if action.r#type == "UpdateConfiguration" {
             let (setting, offset_data) = arguments::decode_argument(&action.body)?;
             let (new_value, _offset_data) = arguments::decode_argument(offset_data)?;
-            let new_value: usize = new_value.parse().map_err(|error| {
-                Status::invalid_argument(format!("New value for {setting} is not valid: {error}"))
-            })?;
+
+            // Parse the new value into None if it is empty and a usize integer if it is not empty.
+            let new_value: Option<usize> = (!new_value.is_empty())
```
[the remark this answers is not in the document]
CGodiksen: I have kept it as it was since it makes it a little easier to tranpose the option and return the error if there is one.

crates/modelardb_server/src/storage/data_transfer.rs
CGodiksen: Good idea, the idea of using a crate was mainly due to initially wanting the scheduling to be more advanced (fx. cron) but since we use fairly simply scheduling it seems fine to use just tokio.

crates/modelardb_server/src/storage/data_transfer.rs
CGodiksen: Fixed.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: Fixed.

crates/modelardb_server/src/storage/mod.rs
CGodiksen: Good idea, fixed.

crates/modelardb_server/tests/integration_test.rs
CGodiksen: I have kept it as it was since `invalid digit found in string` is the error returned from `parse()` and not our message.

skejserjensen (approved):

docs/user/README.md
```diff
@@ -259,9 +259,11 @@ variables is provided here. If an environment variable is not set, the specified
...
 | MODELARDBD_COMPRESSED_RESERVED_MEMORY_IN_BYTES   | 512 MB      | The amount of memory to reserve for storing compressed data buffers.                                                                                                                                                                                                                                     |
 | MODELARDBD_TRANSFER_BATCH_SIZE_IN_BYTES          | 64 MB       | The amount of data that must be collected before transferring a batch of data to the remote object store.                                                                                                                                                                                                |
+| MODELARDBD_TRANSFER_TIME_IN_SECONDS              | Disabled    | The number of seconds between each transfer of data to the remote object store.                                                                                                                                                                                                                          |
```
chrthomsen, line 266: If I understand it correctly, it is possible to set this value to 0 and then the system will always transfer data immediately. Is that correct? Should it be mentioned here?

chrthomsen (approved):

docs/user/README.md
CGodiksen: It is technically correct but I think it is worth looking into edge-cases like this for the configuration so I have created a [work item](https://dev.azure.com/ModelarData/ModelarDB%20Implementation/_workitems/edit/824) to look into this further.

PR #157, opened 2024-01-24
Updated deprecated Docker deployment

CGodiksen:
This PR updates the included deprecated Docker deployment that previously created a single edge deployment that transferred compressed data to a MinIO object store. Since this is no longer possible, the old deployment has been replaced by two new Docker environments. One for a simple local single edge deployment and one for a cluster deployment that includes MinIO, PostgreSQL, a manager node, an edge node, and a cloud node.

The section that describes the Docker environment in the user documentation has also been updated to reflect this change.


docker-compose-cluster.yml
```diff
@@ -0,0 +1,106 @@
...
+      - postgres
+
+  modelardb-edge:
+    image: modelardb
+    container_name: modelardb-edge
+    command: ["target/release/modelardbd", "edge", "data/edge", "grpc://modelardb-manager:9998"]
+    ports:
+      - "9999:9999"
+    environment:
+      MODELARDBD_UNCOMPRESSED_DATA_BUFFER_CAPACITY: 640
```
skejserjensen, line 82: Why overwrite the default settings? If the default settings can be improved I suggest changing them instead.

docs/user/README.md
```diff
@@ -269,31 +269,67 @@ variables is provided here. If an environment variable is not set, the specified
...
-created by running the services defined in [docker-compose-minio.yml](/docker-compose-minio.yml). The services can
-be built and started using the command:
+Two different [Docker](https://docs.docker.com/) environments are included to make it easy to experiment with the 
+different use cases of ModelarDB. The first environment sets up a single instance of `modelardbd` that only uses local 
+storage. Time series data can be ingested into this instance, compressed, and saved to local storage. The compressed 
+data in local storage can then be queried. The second environment covers the more complex use case of ModelarDB where 
+multiple instances of `modelardbd` are deployed in a cluster with a manager that is responsible for managing the 
+cluster. A single edge node and a single cloud node is set up in the cluster. Data can be ingested into the edge or 
+cloud node, compressed, and transferred to a remote object store. The compressed data in the remote object store can 
+then be queried through the cloud node or by directing the query through the manager node.
```
skejserjensen, line 279: Maybe change the description of query processing in the cloud to first say that users can submit the query to the manager, which through workload balancing, assigns the query to a cloud node for execution. As a side note at the end we can write that users also can select a specific cloud node manually, preferably with an example of why they may want to do so, but that querying through the manager for workload balancing is recommended.

docs/user/README.md
```diff
@@ -269,31 +269,67 @@ variables is provided here. If an environment variable is not set, the specified
...
+
+Note that since [Rust](https://www.rust-lang.org/) is a compiled language and a more dynamic ModelarDB configuration
+might be needed, it is not recommended to use the [Docker](https://docs.docker.com/) environments during active 
+development of ModelarDB. They are however ideal to use for experimenting with ModelarDB or when developing 
+components that utilize ModelarDB. Downloading [Docker Desktop](https://docs.docker.com/desktop/) is recommended to
```
skejserjensen, line 284: Replace "components" with "software" as components may be misunderstood as ModelarDB components.

docs/user/README.md
```diff
@@ -269,31 +269,67 @@ variables is provided here. If an environment variable is not set, the specified
...
-The instance can then be accessed using the Apache Arrow Flight interface at `grpc://127.0.0.1:9999`.
+The cluster deployment sets up a [MinIO](https://min.io/) object store and a [MinIO](https://min.io/) client is used to 
+initialize the bucket `modelardb` in the object store. [MinIO](https://min.io/) can be administered through its 
+[web interface](http://127.0.0.1:9001). The default username and password, `minioadmin`, can be used to log in. The 
+deployment also sets up a PostgreSQL database named `metadata` that is used as the metadata database for the cluster. 
+The username `modelardb_user` and the password `modelardb_password` can be used to access the database. 
+
+The cluster itself consists of a manager node, an edge node, and a cloud node. The manager node can be accessed using 
+the URL `grpc://127.0.0.1:9998`, the edge node using the URL `grpc://127.0.0.1:9999`, and the cloud node using the URL
+`grpc://127.0.0.1:9997`. Tables can be created through the manager node and data can be ingested, compressed, and
```
skejserjensen, line 333: Maybe delete the text after `grpc://127.0.0.1:9997` as it seems to repeat what has been written above. If this is done, maybe combine the resulting short paragraph with the paragraph right above it about the cluster deployment setup. If the text is kept and the text above about querying cloud nodes is updated, this text should also be updated for consistency.

skejserjensen (changes requested):

docker-compose-cluster.yml
CGodiksen: The default settings are more ideal for a production build. The values here are the same as were used in the MORE cluster deployment since they better fit a testing/development build.

docs/user/README.md
CGodiksen: I prefer to keep the descriptions very brief in this Docker section and talk about it more detail in the Usage documentation instead (which was updated in the other PR). I will look at those sections again to make sure this is clear. 

This text is also a description of what can be done in the environment that is set up which only includes a single cloud node so talking about workload balancing here does not make much sense,.

docs/user/README.md
CGodiksen: Fixed.

docs/user/README.md
CGodiksen: I have kept it as it was since I think it makes sense to both introduce both deployments in the introduction of the section and also reiterate in the subsections what is possible now that the deployment is running.

skejserjensen (approved):

docs/user/README.md
```diff
@@ -269,31 +269,68 @@ variables is provided here. If an environment variable is not set, the specified
...
+data in local storage can then be queried. The second environment covers the more complex use case of ModelarDB where 
+multiple instances of `modelardbd` are deployed in a cluster with a manager that is responsible for managing the 
+cluster. A single edge node and a single cloud node is set up in the cluster. Data can be ingested into the edge or 
+cloud node, compressed, and transferred to a remote object store. The compressed data in the remote object store can 
+then be queried through the cloud node or by directing the query through the manager node.
+
+Note that since [Rust](https://www.rust-lang.org/) is a compiled language and a more dynamic ModelarDB configuration
+might be needed, it is not recommended to use the [Docker](https://docs.docker.com/) environments during active 
+development of ModelarDB. They are however ideal to use for experimenting with ModelarDB or when developing 
+software that utilize ModelarDB. Downloading [Docker Desktop](https://docs.docker.com/desktop/) is recommended to
```
chrthomsen, line 284: utilizes

chrthomsen (approved):

PR #160, opened 2024-01-29
Prevent timestamp compression from overflowing

skejserjensen:
This PR modifies the Gorilla-based timestamp compression so it store all residual timestamps as a bit prefix which indicates a size and a delta-of-delta. This allows all residual timestamps to be stored using an appropriate number of bites without the possibility of overflow which could happen when using 14 bits for the first residual timestamp as described in the Gorilla paper. In addition it simplifies the code to store the first residual timestamp as a delta-of-delta like all the other timestamp instead of a delta like in the Gorilla paper. To compute a delta-of-delta for the second timestamp a hard coded second delta with the value zero is used.

### DIFF
--- a/crates/modelardb_compression/src/models/bits.rs
+++ b/crates/modelardb_compression/src/models/bits.rs
@@ -53,7 +53,7 @@ impl<'a> BitReader<'a> {
     }
 
     /// Read the next `number_of_bits` bits from the [`BitReader`]. Assumes that
-    /// `number_of_bits` is less than or equal to 32.
+    /// `number_of_bits` is less than or equal to 64.
     pub fn read_bits(&mut self, number_of_bits: u8) -> u64 {
         debug_assert!(
             number_of_bits <= 64,
--- a/crates/modelardb_compression/src/models/bits.rs
+++ b/crates/modelardb_compression/src/models/bits.rs
@@ -54,10 +54,10 @@ impl<'a> BitReader<'a> {
 
     /// Read the next `number_of_bits` bits from the [`BitReader`]. Assumes that
     /// `number_of_bits` is less than or equal to 32.
-    pub fn read_bits(&mut self, number_of_bits: u8) -> u32 {
+    pub fn read_bits(&mut self, number_of_bits: u8) -> u64 {
         debug_assert!(
-            number_of_bits <= 32,
-            "The number of bits to read must be less than or equal to 32."
+            number_of_bits <= 64,
+            "The number of bits to read must be less than or equal to 64."
         );
 
         let mut value = 0;
@@ -74,7 +74,7 @@ impl<'a> BitReader<'a> {
             value = (value << 1) | bit;
         }
         self.next_bit = end_bit;
-        value as u32
+        value
     }
 }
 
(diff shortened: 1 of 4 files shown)


chrthomsen (changes requested):

crates/modelardb_compression/src/models/bits.rs
```diff
@@ -54,10 +54,10 @@ impl<'a> BitReader<'a> {
 
     /// Read the next `number_of_bits` bits from the [`BitReader`]. Assumes that
     /// `number_of_bits` is less than or equal to 32.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_compression/src/models/timestamps.rs
```diff
@@ -107,35 +107,35 @@ fn compress_irregular_residual_timestamps(uncompressed_timestamps: &[Timestamp])
...
+    // computed and then encoded in buckets. Zero is used as the delta for the first delta-of-delta.
+    let mut last_timestamp = uncompressed_timestamps[0];
+    let mut last_delta = 0;
+    for timestamp in &uncompressed_timestamps[1..uncompressed_timestamps.len() - 1] {
         let delta = timestamp - last_timestamp;
         let delta_of_delta = delta - last_delta;
 
         match delta_of_delta {
             0 => compressed_timestamps.append_a_zero_bit(),
             -63..=64 => {
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_compression/src/models/timestamps.rs
```diff
@@ -249,7 +251,7 @@ fn decompress_all_irregular_timestamps(
         }
 
         let delta = match leading_one_bits {
-            0 => last_delta,                                               // Flag is 0.
+            0 => last_delta,                                                             // Flag is 0.
```
chrthomsen, line 254: tiny detail: why extra spaces here?

chrthomsen (approved):

crates/modelardb_compression/src/models/timestamps.rs
skejserjensen: Fixed.

crates/modelardb_compression/src/models/timestamps.rs
```diff
@@ -327,17 +331,93 @@ mod tests {
             1579694400353,
             1579694400493,
             1579694400650,
-        ]);
+        ], Some(4));
+    }
+
+    #[test]
+    fn compress_and_decompress_bucket_sized_1_timestamps_for_an_irregular_time_series() {
```
CGodiksen, line 338: I am not sure this is an universal convention in Rust but it seems that many of our other tests always have "test_" as a prefix in the test function name. If this is a convention we want to use then it should be changed in the tests in this file.

crates/modelardb_compression/src/models/timestamps.rs
```diff
@@ -327,17 +331,93 @@ mod tests {
...
+        compress_and_decompress_timestamps_for_a_time_series(&[
+            500,
+            245, // -255
+            246, //  256
+            500
+        ], Some(4));
+    }
+
+    #[test]
+    fn compress_and_decompress_fourth_bucket_12_sized_timestamps_for_an_irregular_time_series() {
```
CGodiksen, line 367: Very minor but is the naming of this test and the next consistent with the others? Should it be `sized_12` instead.

CGodiksen (changes requested):

crates/modelardb_compression/src/models/timestamps.rs
skejserjensen: Good catch, I have fixed it for all of the tests in this file.

CGodiksen (approved):

PR #162, opened 2024-02-18
Now always reading the full Apache Parquet file when reading record batches

CGodiksen:
This PR fixes a bug that caused Apache Parquet files to not always be read fully. We now loop over the batches in the file until the entire file is read. Duplicated code between `merge_compressed_apache_parquet_files` and `read_batch_from_apache_parquet_file` has been removed using a utility function and the read/write functions and the utility functions have been moved to `modelardb_common` so they can be used everywhere (including Library). 

Since the read/write Parquet functions have been moved, the storage engine now no longer has any associated functions and now only has methods.

It should be noted that reading batches in the Apache Parquet files using concurrency was researched but due to issues with the batches getting read out of order and therefore resulting in an out of order combined record batch, it was dropped.

### DIFF
--- a/crates/modelardb_common/src/storage.rs
+++ b/crates/modelardb_common/src/storage.rs
@@ -15,9 +15,13 @@
 
 //! Utility functions to support reading and writing Apache Parquet files to disk.
 
+use std::ffi::OsStr;
+use std::fs::File;
 use std::io::Write;
+use std::path::Path;
 
 use arrow::array::RecordBatch;
+use arrow::compute;
 use arrow::datatypes::SchemaRef;
 use datafusion::parquet::arrow::async_reader::{AsyncFileReader, ParquetRecordBatchStream};
 use datafusion::parquet::arrow::{ArrowWriter, ParquetRecordBatchStreamBuilder};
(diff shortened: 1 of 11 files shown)


crates/modelardb_common/src/storage.rs
```diff
@@ -0,0 +1,222 @@
...
+    async fn test_read_from_non_existent_path() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let path = temp_dir.path().join("none.parquet");
+        let result = read_batch_from_apache_parquet_file(path.as_path());
+
+        assert!(result.await.is_err());
+    }
+
+    /// Create an Apache Parquet file in the [`tempfile::TempDir`] from a generated [`RecordBatch`].
+    fn create_apache_parquet_file_in_temp_dir(
```
skejserjensen, line 211: Maybe also run the tests that should successfully read `RecordBatch`es with Apache Parquet files that are designed to have multiple `Row Groups` and thus would have failed without the changes in this PR, i. e., specifically adding tests that fails with the old code.

crates/modelardb_common/src/storage.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    #[tokio::test]
+    async fn test_read_entire_apache_parquet_file() {
+        let file_name = "test.parquet".to_owned();
+        let (_temp_dir, path, batch) = create_apache_parquet_file_in_temp_dir(file_name);
+
+        let result = read_batch_from_apache_parquet_file(path.as_path()).await;
+
+        assert!(result.is_ok());
+        assert_eq!(batch, result.unwrap());
```
skejserjensen, line 187: Maybe verify that the correct number of rows where read from the Apache Parquet file to ensure the PR fixes the issue.

crates/modelardb_server/src/context.rs
```diff
@@ -116,9 +117,7 @@ impl Context {
         let session_runtime = Arc::new(RuntimeEnv::default());
 
         // unwrap() is safe as storage::QUERY_DATA_FOLDER_SCHEME_WITH_HOST is a const containing an URL.
```
skejserjensen, line 119: Maybe remove `storage` from the comment to match the code.

skejserjensen (changes requested):

crates/modelardb_common/src/storage.rs
CGodiksen: Fixed.

crates/modelardb_server/src/context.rs
CGodiksen: Fixed.

crates/modelardb_common/src/storage.rs
CGodiksen: Fixed.

skejserjensen (approved):

crates/modelardb_common/src/storage.rs
```diff
@@ -0,0 +1,237 @@
+/* Copyright 2023 The ModelarDB Contributors
```
chrthomsen, line 1: tiny detail, but why not 2024?

chrthomsen (approved):

PR #168, opened 2024-03-20
Add support for Swing's recording mechanism for minimum MSE

aabduvakhobov:
Implementation of recording mechanism for Swing (Section 3.2 in Swing and Slide paper) to minimize MSE. The mechanism generated slightly lower Mean Squared Error and Mean Absolute Percentage Error than the current approach when tested with real-life datasets. However, its improvement over the current approach is not absolute and it varies depending on data points' properties.

### DIFF
--- a/crates/modelardb_compression/src/models/swing.rs
+++ b/crates/modelardb_compression/src/models/swing.rs
@@ -90,8 +90,8 @@ impl Swing {
     /// - (2) When the second data point is received, two linear functions that
     /// intersect with the first data point are computed to designate the upper
     /// and lower bounds for the linear functions Swing can fit to the segment.
-    /// From the second data point, the recording mechanism that is used to
-    /// find the slope with minimum MSE is updated.    
+    /// From the second data point, the nominator and denominator of
+    /// the slope with minimum MSE is updated.    
     /// - (3) Then for each subsequent data point, Swing determines if the data
     /// point can be represented by a linear function in the space delimited by
     /// the upper and lower bounds and updates these bounds if necessary.
@@ -183,7 +183,7 @@ impl Swing {
                 let (
                     current_line_slope_with_minimum_mse_numerator,
                     current_line_slope_with_minimum_mse_denominator,
-                ) = compute_slope_with_minimum_mse(
+                ) = self.compute_slope_with_minimum_mse(
                     self.start_time,
                     self.first_value,
                     timestamp,
@@ -196,6 +196,36 @@ impl Swing {
                 self.length += 1;
                 true
             }
+        }
+    }
+
+    /// Compute the slope that minimizes the MSE for the current line segment based on
+    /// the first value (`start_time`, `first_value`) and (`end_time`, `last_value`).
+    /// Return numerator and denominator of the best slope.
+    ///
+    /// After the second data point, Swing starts computing the best slope for
+    /// each current line segment. Once the data point that exceeds the error bound arrives,
+    /// Swing compares the summation of the slopes with minimum MSE to the upper and lower
+    /// bounds to find the best slope.
+    ///
+    /// For more detail see Equations 1-6 in the [Swing and Slide paper].
+    ///
+    /// [Swing and Slide paper]: https://dl.acm.org/doi/10.14778/1687627.1687645
+    fn compute_slope_with_minimum_mse(
+        &mut self,
+        start_time: Timestamp,
+        first_value: f64,
+        end_time: Timestamp,
+        last_value: f64,
+    ) -> (f64, f64) {
+        if models::equal_or_nan(first_value, last_value) {
+            (0.0, 0.0)
+        } else {
+            debug_assert!(first_value.is_finite(), "First value is not finite.");
+            debug_assert!(last_value.is_finite(), "Last value is not finite.");
+            let numerator = (last_value - first_value) * (end_time - start_time) as f64;
+            let denominator = ((end_time - start_time) as f64).powi(2);
+            (numerator, denominator)
         }
     }
 
@@ -218,9 +248,10 @@ impl Swing {
     /// must be [`f64`] to be precise enough.
     pub fn model(self) -> (Value, Value) {
         // Implementation of the Equations for minimizing MSE as specified in the Swing and Slide paper.
-        // Equations 5 and 6 in the Swing and Slide paper.
+        // Equation 6 in the Swing and Slide paper.
         let projected_slope =
             self.slope_with_minimum_mse_numerator / self.slope_with_minimum_mse_denominator;
+        // Equation 5 in the Swing and Slide paper.
         let slope_with_minimum_mse = self
             .lower_bound_slope
             .max(projected_slope.min(self.upper_bound_slope));
@@ -228,35 +259,6 @@ impl Swing {
         let last_value =
             slope_with_minimum_mse * (self.end_time - self.start_time) as f64 + self.first_value;
         (self.first_value as Value, last_value as Value)
-    }
-}
-
-/// Compute the slope that minimizes the MSE for the current line segment based on
-/// the first value (`start_time`, `first_value`) and (`end_time`, `last_value`).
-/// Return numerator and denominator of the best slope.
-///
-/// After the second data point, Swing starts computing the best slope for
-/// each current line segment. Once the data point that exceeds the error bound arrives,
-/// Swing compares the summation of the slopes with minimum MSE to the upper and lower
-/// bounds to find the best slope.
-///
-/// For more detail see Equations 1-6 in the [Swing and Slide paper].
-///
-/// [Swing and Slide paper]: https://dl.acm.org/doi/10.14778/1687627.1687645
-fn compute_slope_with_minimum_mse(
-    start_time: Timestamp,
-    first_value: f64,
-    end_time: Timestamp,
-    last_value: f64,
-) -> (f64, f64) {
-    if models::equal_or_nan(first_value, last_value) {
-        (0.0, 0.0)
-    } else {
-        debug_assert!(first_value.is_finite(), "First value is not finite.");
-        debug_assert!(last_value.is_finite(), "Last value is not finite.");
-        let numerator = (last_value - first_value) * (end_time - start_time) as f64;
-        let denominator = ((end_time - start_time) as f64).powi(2);
-        (numerator, denominator)
     }
 }
 
@@ -685,17 +687,18 @@ mod tests {
             &[42.0, 42.0, 42.8, 42.0, 42.0],
         ))
     }
-    // #[test]
-    // fn test_can_minimize_mse_with_relative_error_bound_five() {
-    //     assert!(can_minimize_mse(
-    //         ErrorBound::try_new_relative(ERROR_BOUND_FIVE).unwrap(),
-    //         &[42.0, 42.0, 42.8, 42.0, 41.0, 40.0, 42.0, 42.0, 42.0, 42.1]
-    //     ))
-    // }
 
     #[test]
     fn test_slope_is_between_higher_and_lower_hyperplane() {
         assert!(slope_is_between_higher_lower_hyperplanes(
+            ErrorBound::try_new_relative(ERROR_BOUND_FIVE).unwrap(),
+            &[42.0, 42.0, 42.8, 42.0, 41.0, 40.0, 42.0, 42.0, 42.0, 42.1]
+        ))
+    }
+
+    #[test]
+    fn test_can_minimize_mse_with_relative_error_bound_five() {
+        assert!(can_minimize_mse(
             ErrorBound::try_new_relative(ERROR_BOUND_FIVE).unwrap(),
             &[42.0, 42.0, 42.8, 42.0, 41.0, 40.0, 42.0, 42.0, 42.0, 42.1]
         ))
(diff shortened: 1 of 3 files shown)


crates/modelardb_compression/src/models/swing.rs
```diff
@@ -56,6 +56,10 @@ pub struct Swing {
     /// Intercept for the linear function specifying the lower bound for the
     /// current model.
     lower_bound_intercept: f64,
+    /// Computation of the slope to minimize MSE for the current model 
+    /// and holds the numenator and denominator of Equation 6 
+    /// described in Section 3.2 of [Swing and Slide paper].
```
CGodiksen, line 61: I am not sure this link works and actually links to the paper.  `[Swing and Slide paper](https://dl.acm.org/doi/10.14778/1687627.1687645)` can be used to link to the paper.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -56,6 +56,10 @@ pub struct Swing {
     /// Intercept for the linear function specifying the lower bound for the
     /// current model.
     lower_bound_intercept: f64,
+    /// Computation of the slope to minimize MSE for the current model 
+    /// and holds the numenator and denominator of Equation 6
```
CGodiksen, line 60: I am not sure this sentence makes sense. Maybe `Computation of the slope to minimize MSE for the current model. Specifically, it is the numerator and denominator...` or something similar instead.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -169,6 +177,18 @@ impl Swing {
                             value - maximum_deviation,
                         );
                 }
+        
+                // Incremental computation of Equation 6 in Section 3.2 in the Swing and Slide paper
```
CGodiksen, line 181: Missing punctuation.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -193,12 +213,13 @@ impl Swing {
...
     /// must be [`f64`] to be precise enough.
     pub fn model(self) -> (Value, Value) {
-        // TODO: use the function with the minimum error as specified in the Swing and Slide paper.
-        let average_slope = (self.lower_bound_slope + self.upper_bound_slope) / 2.0;
-        let average_intercept = (self.lower_bound_intercept + self.upper_bound_intercept) / 2.0;
-        let first_value = average_slope * self.start_time as f64 + average_intercept;
-        let last_value = average_slope * self.end_time as f64 + average_intercept;
-        (first_value as Value, last_value as Value)
+        // Implementation of the function with the minimized MSE as specified in the Swing and Slide paper.
+        // Equation 5 and 6
```
CGodiksen, line 217: Maybe add `in the Swing and Slide paper.`.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -193,12 +213,13 @@ impl Swing {
...
+        let projected_slope = self.slope_with_minimum_mse.0 / self.slope_with_minimum_mse.1;
+        let optimal_slope =  self.lower_bound_slope.max(projected_slope.min(self.upper_bound_slope));
+        // Equation 2
```
CGodiksen, line 220: Maybe add `in the Swing and Slide paper.`.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -193,12 +213,13 @@ impl Swing {
...
-        // TODO: use the function with the minimum error as specified in the Swing and Slide paper.
-        let average_slope = (self.lower_bound_slope + self.upper_bound_slope) / 2.0;
-        let average_intercept = (self.lower_bound_intercept + self.upper_bound_intercept) / 2.0;
-        let first_value = average_slope * self.start_time as f64 + average_intercept;
-        let last_value = average_slope * self.end_time as f64 + average_intercept;
-        (first_value as Value, last_value as Value)
+        // Implementation of the function with the minimized MSE as specified in the Swing and Slide paper.
+        // Equation 5 and 6
+        let projected_slope = self.slope_with_minimum_mse.0 / self.slope_with_minimum_mse.1;
+        let optimal_slope =  self.lower_bound_slope.max(projected_slope.min(self.upper_bound_slope));
```
CGodiksen, line 219: It seems there is an extra space after `=`. Maybe use `Rustfmt` to be sure all issues like this is fixed.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -326,6 +347,25 @@ fn compute_slope_and_intercept(
     }
 }
 
+/// Computes numenator and denominator for Equation 6 in Section 3.2 of Swing and Slide paper.
```
CGodiksen, line 350: `numerator` instead of `numenator`.

crates/modelardb_compression/src/models/swing.rs
CGodiksen, line 60: `numerator` instead of `numenator`.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -326,6 +347,25 @@ fn compute_slope_and_intercept(
...
+    first_value: f64,
+    end_time: Timestamp,
+    last_value: f64,
+) -> (f64, f64) {
+    if first_value == last_value || (first_value.is_nan() || last_value.is_nan()) {
+        (0.0, 0.0)
+    } else {
+        debug_assert!(first_value.is_finite(), "First value is not finite.");
+        debug_assert!(last_value.is_finite(), "Last value is not finite.");
+        let numenator =  (last_value - first_value) * (end_time - start_time) as f64;
```
CGodiksen, line 363: `numerator` instead of `numenator`.

crates/modelardb_compression/src/models/swing.rs
CGodiksen, line 350: Maybe add link to paper as in doc comment for `fn fit_data_point`.

CGodiksen (changes requested):

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed. Removed the link since it is given above.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

CGodiksen (approved):

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -84,9 +89,12 @@ impl Swing {
     /// - (2) When the second data point is received, two linear functions that
     /// intersect with the first data point are computed to designate the upper
     /// and lower bounds for the linear functions Swing can fit to the segment.
+    /// From the second data point, the recording mechanism that is used to
```
skejserjensen, line 92: Unclear what "the recording mechanism is updated" mean, do you mean minimum slope is updated? Maybe simplify doc comment.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -169,6 +177,18 @@ impl Swing {
                             value - maximum_deviation,
                         );
                 }
+
+                // Incremental computation of Equation 6 in Section 3.2 in the Swing and Slide paper.
```
skejserjensen, line 181: See my comment above about computing the slope with minimum MSE once in `model(self)` except of once per data point.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -326,6 +350,26 @@ fn compute_slope_and_intercept(
     }
 }
 
+/// Computes numerator and denominator for Equation 6
```
skejserjensen, line 353: Maybe update the doc comments for the function so they describe what the function does like is done for the other functions. The doc comment should be self-contained enough that the reader can understand the purpose of the function, e.g., see `fit_data_pont()`.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -326,6 +350,26 @@ fn compute_slope_and_intercept(
     }
 }
 
+/// Computes numerator and denominator for Equation 6 
+/// in Section 3.2 of [Swing and Slide paper](https://dl.acm.org/doi/10.14778/1687627.1687645).
+/// It is a part of recording mechanism for finding the optimal slope with minimum MSE.
+fn compute_local_optimal_slope(
```
skejserjensen, line 356: Private functions and methods should be placed right after the last function or method that calls them, e.g., see `compute_slope_and_intercept()`. In this case the last caller is `compute_local_optimal_slope()`. If multiple private functions or methods are placed after a function or method they should be placed in the order they are called. So the one that is called first should be placed first, followed by the one that is called second, and so on.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -326,6 +350,26 @@ fn compute_slope_and_intercept(
...
+    start_time: Timestamp,
+    first_value: f64,
+    end_time: Timestamp,
+    last_value: f64,
+) -> (f64, f64) {
+    if first_value == last_value || (first_value.is_nan() || last_value.is_nan()) {
```
skejserjensen, line 362: Unclear why this if expressions check if either of the values are `NaN` instead of both like `compute_slope_and_intercept()`. Maybe add a comment stating why or change the code to use `models::equal_or_nan()` if it should check if both values are `NaN`.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -326,6 +350,26 @@ fn compute_slope_and_intercept(
...
+        debug_assert!(first_value.is_finite(), "First value is not finite.");
+        debug_assert!(last_value.is_finite(), "Last value is not finite.");
+        let numerator = (last_value - first_value) * (end_time - start_time) as f64;
+        let denominator = ((end_time - start_time) as f64).powi(2);
+        (numerator, denominator)
+    }
+}
+
 #[cfg(test)]
 mod tests {
```
skejserjensen, line 374: Unit tests should be added to verify that the code added in this PR makes Swing produce linear functions with minimum MSE. The unit tests should test all aspects of the new code, e.g., by compression sequences of data points that Swing would previously not fit a linear function to with an optimal MSE. If possibly, tests that use data points generated using `proptest` or `modelardb_common::test::data_generation` should be added to improve our confidence that the code is correct and to catch bugs that happen due to changes in the code elsewhere. Remember to check that the tests actually tests the effect of the code added by this PR, e.g., that the tests fail if `compute_local_optimal_slope()` is not called.

crates/modelardb_compression/src/models/swing.rs
skejserjensen, line 356: Unclear what `local` and `optimal` means in this context, optimal can almost always be replaced with a more precise term. For example, in this case is `optimal` the lowest or highest value and for what metric? Maybe rename the function to more clearly state what it computes.

skejserjensen (changes requested):

skejserjensen (changes requested):
There are multiple errors reported by [GitHub Actions](https://github.com/ModelarData/ModelarDB-RS/actions/runs/8423249416/job/23097152463?pr=168) that needs to be fixed before the next round of reviews. To automatically run all of the checks in the current [workflow](https://github.com/ModelarData/ModelarDB-RS/blob/main/.github/workflows/cargo-build-lint-and-test-on-pr-and-push.yml), this [Git hook](https://github.com/ModelarData/Utilities/blob/main/Git-Hooks/rust-pre-commit) can be used with the latest version of Rust installed by [Rustup](https://rustup.rs/). Also please provide an answer to each of the comments that was given by the reviewer, currently there are none as can be seen below. 20240326_18h53m23s_grim If it is trivial to understand what code was changed to handle a particular comments a single "Fixed." or similar is fine, although, if the changes are complex please provide an answer that makes it clear what changes was made to handle the comment.

skejserjensen:
> For the last commit, the rust on Ubuntu was updated to _rustc 1.77.0 (aedd173a2 2024-03-17)_.
> 
> There was one issue in the `crates/modelardb_compression/src/models/swing.rs` file that was fixed, but running `cargo clippy` or the Git hook you shared still finds 5 errors in completely another part of the code: `crates/modelardb_server/src/optimizer/model_simple_aggregates.rs`. You could check `cargo clippy` error messages [GitHub Actions](https://github.com/ModelarData/ModelarDB-RS/actions/runs/8423249416/job/23097152463?pr=168).

Although the `clippy` errors are not caused by this PR, they still have to be fixed in order for this PR to be merged as branch protection requires all checks to complete without errors or warnings.


> Regarding the replies to your comments, I replied to all of them (most of the comments say "Fixed"), but strangely they are still pending on my side. Removing and adding a new one does not help.

Unfortunately I still cannot see your comments and when doing code review it is the person that creates the pending comments that have to make them visible by submitting the review. From looking at Google results for "_github pending comments_", it looks like it is possible to create comments as `pending` so the author of the comment has to submit them afterwards, see, e.g., this [discussion](https://github.com/orgs/community/discussions/10369) and [issue](https://github.com/microsoft/vscode-pull-request-github/issues/4195).

aabduvakhobov:
I could also submit my comments. Thanks for the hint!

aabduvakhobov:
After running couple of times, the test `test_try_compress_irregular_synthetic_time_series_without_noise_within_absolute_error_bound_five` in  `compression.rs`  fails for the Swing model even with its initial implementation. Should this be indicated as an issue?

image

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -193,12 +217,46 @@ impl Swing {
...
     /// must be [`f64`] to be precise enough.
     pub fn model(self) -> (Value, Value) {
-        // TODO: use the function with the minimum error as specified in the Swing and Slide paper.
-        let average_slope = (self.lower_bound_slope + self.upper_bound_slope) / 2.0;
-        let average_intercept = (self.lower_bound_intercept + self.upper_bound_intercept) / 2.0;
-        let first_value = average_slope * self.start_time as f64 + average_intercept;
-        let last_value = average_slope * self.end_time as f64 + average_intercept;
-        (first_value as Value, last_value as Value)
+        // Implementation of the Equations for minimizing MSE as specified in the Swing and Slide paper.
+        // Equations 5 and 6 in the Swing and Slide paper.
```
skejserjensen, line 221: Maybe split the comment into two and add it to each line to make it clear which equation the specific code matches.

crates/modelardb_compression/src/models/swing.rs
skejserjensen: It is still unclear what "recording mechanism" means in this context, can the comment be rewritten so it is removed?

crates/modelardb_compression/src/models/swing.rs
skejserjensen: Since `fit_data_point()` is the last caller of `compute_slope_with_minimum_mse()`, `compute_slope_with_minimum_mse()` should be placed right after `fit_data_point()`.

crates/modelardb_compression/src/models/swing.rs
skejserjensen: > "e.g., by compression sequences of data points that Swing would previously not fit a linear function to with an optimal MSE " the length of approximated sequences doesn't increase as a result of this change.

Yes the length should be the same but the MSE should improve as a result of these changes. Thus, it would be good if tests were added that verify that the models that are now being produced now has a lower MSE. For example, by using a sequence of data points that Swing compressed with a linear function with a higher MSE before these changes were added so the test fails without these changes. The best case would be to automatically generate random data and then check that the MSE is always smaller than before, however, I do not know if that is possible without keeping the old version as well. Thus, using a sequence of data points that now has a reduced MSE is probably the more reasonable approach despite not testing the code as extensively.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -627,6 +685,21 @@ mod tests {
             &[42.0, 42.0, 42.8, 42.0, 42.0],
         ))
     }
+    // #[test]
```
skejserjensen, line 688: Leftover code should not be merged into the `main` branch. If code is no longer used it should be deleted. However, in this case it looks like unfinished tests, so it would probably be better to finish implementing the tests instead of just deleting the code.

crates/modelardb_server/src/main.rs
```diff
@@ -15,7 +15,7 @@
 
 //! Implementation of ModelarDB's main function.
 
-#![allow(clippy::too_many_arguments)]
+#![allow(clippy::too_many_arguments, clippy::unconditional_recursion)]
```
skejserjensen, line 18: Clippy warnings should be fixed if at all possible. If it is not possible at all, the place in the code that emit that warning should be annotated with `clippy::unconditional_recursion` so it is clear what part of the code the warning is suppressed from. By placing `clippy::unconditional_recursion` here, all `clippy::unconditional_recursion` warnings are suppressed.

skejserjensen (changes requested):

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: The test was added.

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

crates/modelardb_server/src/main.rs
aabduvakhobov: As I mentioned fixing it required considerable changes, so I placed `clippy::unconditional_recursion` in the respective file.

crates/modelardb_compression/src/models/swing.rs
```diff
@@ -712,49 +715,50 @@ mod tests {
...
-    //         mse += (values[i] as f64 - approximate_value).powi(2);
-    //         let approximate_value_with_minimized_mse =
-    //             slope_with_minimum_mse * timestamp as f64 + intercept_with_minimum_mse;
-    //         optimized_mse += (values[i] as f64 - approximate_value_with_minimized_mse).powi(2);
-    //     }
-    //     let mse = mse / values.len() as f64;
-    //     let optimized_mse = optimized_mse / values.len() as f64;
-    //     mse > optimized_mse
-    // }
+    fn can_minimize_mse(error_bound: ErrorBound, values: &[Value]) -> bool {
```
skejserjensen, line 718: Very minor so I accepted the PR and asked @chrthomsen for the final round of review, but the missing punctuation at the end of the comments.

crates/modelardb_server/src/main.rs
skejserjensen: Ok, I have added a task to fix it to the backlog.

skejserjensen:
> After running couple of times, the test `test_try_compress_irregular_synthetic_time_series_without_noise_within_absolute_error_bound_five` in `compression.rs` fails for the Swing model even with its initial implementation. Should this be indicated as an issue?
> 
> image

It is a known issue the Swing currently can deviate slightly from the error bound if extreme values are compressed. This of course has to be fixed before ModelarDB can be deployed for any real-life use-cases. Regarding if this should be added as an issue, unfortunately, the project currently manages issues through Azure DevOps as it was the task management software we where most familiar with when we started developing the Rust-based version of ModelarDB. However, since ModelarDB is being developed as an open-source project the backlog, issues, etc should be public to enable collaboration and make it clear which issues are known. Thus, I am currently looking into migrating the information in Azure DevOps to GitHub as either a GitHub Project or as GitHub Issues. Ideas for how to do this and/or manage projects using the tools provided by GitHub is welcome.

skejserjensen (approved):
@aabduvakhobov, thank you very much for creating the PR and fixing the issues raised. I now think this PR looks good enough to be merged. A few punctuations are missing from a few comments, but this is very very minor and is easily fixed as part of the changes requested by @chrthomsen's review.

aabduvakhobov:
Thank you all for your very nice guidance! Hope to contribute much more :)

crates/modelardb_compression/src/models/swing.rs
aabduvakhobov: Fixed.

chrthomsen (approved):
Thanks and congrats on your first contribution to ModelarDB-RS @aabduvakhobov

PR #169, opened 2024-04-16
Add custom syntax checking using ast-grep

skejserjensen:
This PR adds custom rules for syntax checking of the code using [`ast-grep`](https://ast-grep.github.io/) and makes GitHub Actions automatic use these rules to check all PRs and the main branch.

### DIFF
--- a/rules/leftover_debugging.yml
+++ b/rules/leftover_debugging.yml
@@ -1,0 +1,6 @@
+id: leftover_debugging
+message: Leftover debug constructs or inforamtion.
+severity: error
+language: Rust
+rule:
+  pattern: dbg!($$$_)
--- a/rules/leftover_debugging.yml
+++ b/rules/leftover_debugging.yml
@@ -1,5 +1,5 @@
 id: leftover_debugging
-message: Leftover debug constructs or inforamtion.
+message: Leftover debug constructs or information.
 severity: error
 language: Rust
 rule:
--- a/rules/missing_punctuation.yml
+++ b/rules/missing_punctuation.yml
@@ -1,0 +1,14 @@
+id: missing_punctuation
+message: Comment is missing punctuation.
+severity: error
+language: Rust
+# The last line of a comment must end with punctuation unless it is a link.
+rule:
+  kind: line_comment
+  all:
+    - regex: '.+[^\\.]$'
+    - not:
+        regex: '.*https?://.*'
+  precedes:
+    not:
+      kind: line_comment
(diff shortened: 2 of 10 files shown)


rules/leftover_debugging.yml
```diff
@@ -0,0 +1,6 @@
+id: leftover_debugging
+message: Leftover debug constructs or inforamtion.
```
CGodiksen, line 2: "information" instead.

rules/missing_punctuation.yml
```diff
@@ -0,0 +1,14 @@
+id: missing_punctuation
```
CGodiksen, line 1: A rule could also be added to do this for log messages and error messages.

rules/test_not_named_test.yml
```diff
@@ -0,0 +1,14 @@
+id: test_not_named_test
+message: Test function not named test_.
+severity: error
+language: Rust
+# Test functions (annotated with #[test] or #[tokio::test]) be named test_ etc.
```
CGodiksen, line 5: Maybe `must be named test_ etc.` instead.

rules/missing_punctuation.yml
CGodiksen, line 1: We could also consider rules for when doc comments are necessary. Currently all modules, structs, struct fields, methods (except `new()`), and functions have doc comments. 

Note that this rule might not be ideal since we might have very small functions and test utility functions that we do not think need doc comments.

CGodiksen, line 1: I think it would be hard to enforce the grouped structure of our imports with a rule in general but maybe we could add a rule to make sure `use super::*;` is always first in the test imports and always followed by an empty line. Quickly looking through the code it seems we are a little inconsistent here.

CGodiksen (changes requested):

rules/leftover_debugging.yml
skejserjensen: Fixed.

rules/missing_punctuation.yml
skejserjensen: Add to task.

rules/test_not_named_test.yml
skejserjensen: Fixed.

rules/missing_punctuation.yml
skejserjensen: Added to task.

skejserjensen: Added to task.

CGodiksen (approved):

chrthomsen (approved):

PR #171, opened 2024-04-29
Add conservative rule that checks for extra space

skejserjensen:
This PR adds an Ast-Grep rule that checks if an extra space has been added to a comment or string. The rule does purposely not match an arbitrary number of space so false positives are not emitted when whitespace is used for formatting. The rule is also limited to comments and string as `rustfmt` removes extra whitespace in the code except for comments and strings. If there are other AST nodes where `rustfmt` does not remove extra whitespace they should be added to this rule.

### DIFF
--- a/rules/extra_space_character.yml
+++ b/rules/extra_space_character.yml
@@ -1,0 +1,10 @@
+id: extra_space_character
+message: Comment or string contains an extra space.
+severity: error
+language: Rust
+# The rule only check for double space to not report errors when whitespace is used for formatting.
+rule:
+  any:
+    - kind: line_comment
+    - kind: string_literal
+  regex: "[^ ]  [^ ]"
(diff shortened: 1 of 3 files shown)


rules/extra_space_character.yml
```diff
@@ -0,0 +1,10 @@
+id: extra_space_character
+message: Comment or string contains an extra space.
+severity: error
+language: Rust
+# The rule only check for double space to not report errors when whitespace is used for formatting.
```
CGodiksen, line 5: Very minor but maybe `The rule only checks for double space` instead.

CGodiksen (approved):

rules/extra_space_character.yml
skejserjensen: Fixed.

chrthomsen (approved):

Issue #197, opened 2024-08-08
Evaluate removing uid by storing tags in segments


skejserjensen:
A way to remedy #187 may be to store tags in segments if it does not significantly increase the amount of space and bandwidth used and then remove `univariate_id`. However, this change needs to be be evaluated in-depth first, as the schema for compressed segments would no longer be the same for model tables with tags and and it may increase the amount of storage and bandwidth required. Removing `univariate_id`s has multiple benefits in addition to fixing #187, e.g., it makes each segment self-contained which reduces the complexity of data transfer and query processing and it removes the limit on columns in model table. The problem with each model table having different schema for compressed segments could be solved by adding the model tables compressed segment schema to `ModelTableMetadata`.

skejserjensen:
After it has been determined if this change should be made and [dev/delta-metadata](https://github.com/ModelarData/ModelarDB-RS/tree/dev/delta-metadata) has been merged, the schemas used for the various `RecordBatch`es in ModelarDB should be updated to not use unsigned integers and the number of schemas should be reduced to the fewest possible. While the use of unsigned integers for values like column index may make the semantics more clear, it does not seem to be worth the extra complexity needed to convert back and forth between unsigned integers in memory and signed integers on disk.

PR #198, opened 2024-08-08
Fix #187 by storing univariate_id as int64

### LINKED ISSUE #187: Some univariate_id cannot be written to Delta Lake
Some `u64` `univariate_id`s cause errors when written to Delta Lake as it looks like they are stored as `Int64` instead of `UInt64` despite the field being defined as `Field::new("univariate_id", DataType::UInt64, false)`. For example `tag_one` and/or `tag_two` [previously](https://github.com/ModelarData/ModelarDB-RS/pull/185/commits/479068853db104db8cd14ac070c2dca9cb516afa) used in the integration tests causes the following error to occur when executing the tests: `Failed to flush data in compressed data manager due to: Failed to convert into Arrow schema: Cast error: Can't cast value 11069825858223412227 to type Int64`.

skejserjensen:
This PR fixes #187 by storing `univariate_id` as `Int64Array` instead of `UInt64Array` as the [Delta](https://github.com/delta-io/delta-rs/issues/2175) [Lake](https://github.com/delta-io/delta-rs/issues/1751) [Protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types) does not support unsigned integers and #197 probably will not be resolved for some time. When reviewing this PR, please double check the necessary casts are performed everywhere they are needed. Currently, `UInt64Array` is reinterpreted as `Int64Array` when writing compressed segments to disk and `Int64Array` is reinterpreted as `UInt64Array` when data points are reconstructed from segments and aggregates are computed directly from the segments.

### DIFF
--- a/crates/modelardb_common/src/storage.rs
+++ b/crates/modelardb_common/src/storage.rs
@@ -359,7 +359,7 @@ pub fn univariate_ids_int64_to_uint64(compressed_segments: &RecordBatch) -> Reco
         signed_univariate_ids.unary(|value| u64::from_ne_bytes(value.to_ne_bytes()));
     columns[0] = Arc::new(univariate_ids);
 
-    // unwrap() is safe as columns are constructs to match DISK_COMPRESSED_SCHEMA.
+    // unwrap() is safe as columns are constructs to match QUERY_COMPRESSED_SCHEMA.
     RecordBatch::try_new(QUERY_COMPRESSED_SCHEMA.0.clone(), columns).unwrap()
 }
 
@@ -506,9 +506,28 @@ mod tests {
 
     use arrow::datatypes::{Field, Schema};
     use object_store::local::LocalFileSystem;
+    use proptest::num::u64 as ProptestUnivariateId;
+    use proptest::{prop_assert_eq, proptest};
     use tempfile::TempDir;
 
-    use crate::test;
+    use crate::test::{self, compressed_segments_record_batch_with_time};
+
+    // Tests for univariate_ids_uint64_to_int64() and univariate_ids_int64_to_uint64().
+    proptest! {
+    #[test]
+    fn test_univariate_ids_uint64_to_int64_to_uint64(univariate_id in ProptestUnivariateId::ANY) {
+        let record_batch = compressed_segments_record_batch_with_time(univariate_id, 0, 0.0);
+        let mut expected_record_batch = record_batch.clone();
+        expected_record_batch.remove_column(10);
+
+        let mut record_batches = vec![record_batch.clone()];
+        univariate_ids_uint64_to_int64(&mut record_batches);
+        record_batches[0].remove_column(10);
+        let computed_record_batch = univariate_ids_int64_to_uint64(&record_batches[0]);
+
+        prop_assert_eq!(expected_record_batch, computed_record_batch);
+    }
+    }
 
     // Tests for read_record_batch_from_apache_parquet_file().
     #[tokio::test]
--- a/crates/modelardb_common/src/storage.rs
+++ b/crates/modelardb_common/src/storage.rs
@@ -18,7 +18,7 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use arrow::array::RecordBatch;
+use arrow::array::{Int64Array, RecordBatch, UInt64Array};
 use arrow::compute;
 use arrow::datatypes::{Field, Schema};
 use datafusion::parquet::arrow::async_reader::{
@@ -42,7 +42,9 @@ use url::Url;
 use uuid::Uuid;
 
 use crate::arguments::decode_argument;
-use crate::schemas::{COMPRESSED_SCHEMA, FIELD_COLUMN};
+use crate::schemas::{
+    COMPRESSED_SCHEMA, DISK_COMPRESSED_SCHEMA, FIELD_COLUMN, QUERY_COMPRESSED_SCHEMA,
+};
 
 /// The folder storing compressed data in the data folders.
 const COMPRESSED_DATA_FOLDER: &str = "tables";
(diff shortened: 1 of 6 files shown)


crates/modelardb_common/src/storage.rs
```diff
@@ -330,6 +335,34 @@ pub async fn extract_azure_blob_storage_arguments(
...
+    }
+}
+
+/// Reinterpret the bits used for univariate ids in `compressed_segments` to convert the column from
+/// [`Int64Array`] to [`UInt64Array`] as the Delta Lake Protocol does not support unsigned integers.
+pub fn univariate_ids_int64_to_uint64(compressed_segments: &RecordBatch) -> RecordBatch {
+    let mut columns = compressed_segments.columns().to_vec();
+    let signed_univariate_ids = crate::array!(compressed_segments, 0, Int64Array);
+    let univariate_ids: UInt64Array =
+        signed_univariate_ids.unary(|value| u64::from_ne_bytes(value.to_ne_bytes()));
```
chrthomsen, line 359: Same as above: is it safe to use native endianness? Do we risk to write with one endianness and then later read with another endianess (on the client)?

crates/modelardb_common/src/storage.rs
```diff
@@ -330,6 +335,34 @@ pub async fn extract_azure_blob_storage_arguments(
...
 }
 
+/// Reinterpret the bits used for univariate ids in `compressed_segments` to convert the column from
+/// [`UInt64Array`] to [`Int64Array`] as the Delta Lake Protocol does not support unsigned integers.
+fn univariate_ids_uint64_to_int64(compressed_segments: &mut Vec<RecordBatch>) {
+    for record_batch in compressed_segments {
+        let mut columns = record_batch.columns().to_vec();
+        let univariate_ids = crate::array!(record_batch, 0, UInt64Array);
+        let signed_univariate_ids: Int64Array =
+            univariate_ids.unary(|value| i64::from_ne_bytes(value.to_ne_bytes()));
```
chrthomsen, line 345: is it safe to use native endianness? Do we risk to write with one endianness and then later read with another endianess (on the client)?

skejserjensen: Attempts have been made to ensure ModelarDB works on systems with different endianess, e.g., the [compression purposely uses little-endian](https://github.com/ModelarData/ModelarDB-RS/blob/84c886f07dbd470ff8a5803ca68dee3ea246fe03/crates/modelardb_compression/src/types.rs#L298) to ensure a specific endianess is used for compressed segments regardless of the system's endianess. Little-endiann was chosen as high performance CPUs such as [X86_64 and M-series](https://developer.apple.com/documentation/apple-silicon/porting-your-macos-apps-to-apple-silicon#Address-Architectural-Differences) uses little-endian. Unfortunately, GitHub Actions does not, as far as I am aware, [provide runners for any big-endian systems](https://github.com/actions/runner-images). Thus, ModelarDB has never been tested on anything but little-endian, so I am certain that there are issues we have not considered, e.g., Apache Arrow specifically states that it is [little-endian by default](https://arrow.apache.org/docs/format/Columnar.html#byte-order-endianness).

Regarding the use of native-endianness in the function executed by `unary()`, it was purposely used as [`unary()`](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveArray.html#method.unary) operates on [`ArrowPrimitiveType`](https://docs.rs/arrow/latest/arrow/datatypes/trait.ArrowPrimitiveType.html) which in this case is [`UInt64Type`](https://docs.rs/arrow/latest/arrow/datatypes/struct.UInt64Type.html) which corresponds to the native Rust type [`u64`](https://doc.rust-lang.org/nightly/std/primitive.u64.html) which is native-endian, e.g., see the documentation for [from_be()](https://doc.rust-lang.org/nightly/std/primitive.u64.html#method.from_be). However, according to the [documentation for Apache Parquet](https://parquet.apache.org/docs/file-format/data-pages/encodings/), it always stores integers as little-endian. So if I understand the process correctly, integers are stored in Apache Parquet as little-endian, read into and stored in Apache Arrow as little-endian, converted to native-endian when manipulated by `unary()` which is a [no-op for little-endian systems](https://doc.rust-lang.org/nightly/std/primitive.u64.html#method.from_le), and than written back to Apache Arrow as little-endian. Thus, the function executed by `unary()` must use native-endianess to be portable. However, there may certainly be something that I have overlooked or misunderstood, and as stated above, the code has only been tested on little-endian systems as I do not have access to systems with any other endianess.

crates/modelardb_common/src/storage.rs
skejserjensen: See [this comment](https://github.com/ModelarData/ModelarDB-RS/pull/198#discussion_r1711403886).

chrthomsen (approved):

crates/modelardb_common/src/storage.rs
```diff
@@ -330,6 +335,34 @@ pub async fn extract_azure_blob_storage_arguments(
...
+        columns[0] = Arc::new(signed_univariate_ids);
+
+        // unwrap() is safe as columns are constructs to match DISK_COMPRESSED_SCHEMA.
```
CGodiksen, line 348: Should this be `as columns is constructed`?

crates/modelardb_common/src/storage.rs
```diff
@@ -330,6 +335,34 @@ pub async fn extract_azure_blob_storage_arguments(
...
+        *record_batch = RecordBatch::try_new(DISK_COMPRESSED_SCHEMA.0.clone(), columns).unwrap();
+    }
+}
+
+/// Reinterpret the bits used for univariate ids in `compressed_segments` to convert the column from
```
CGodiksen, line 353: Maybe make it clear that a new record batch is returned to avoid confusion with the above function that looks similar but instead converts the given record batch directly.

crates/modelardb_common/src/storage.rs
CGodiksen: if this is changed here also change it in the function below.

CGodiksen (changes requested):

crates/modelardb_common/src/storage.rs
skejserjensen: Fixed.

crates/modelardb_common/src/storage.rs
skejserjensen: Fixed.

CGodiksen (approved):

PR #203, opened 2024-08-15
Automatically create the local data folder

skejserjensen:
This PR will make `modelardbd` automatically create the local data folder if it does not exist instead of returning an error.

### DIFF
--- a/crates/modelardb_common/src/storage.rs
+++ b/crates/modelardb_common/src/storage.rs
@@ -66,8 +66,8 @@ pub struct DeltaLake {
 }
 
 impl DeltaLake {
-    /// Create a new [`DeltaLake`] that manages the delta tables in `data_folder_path` if it exists.
-    /// Returns a [`DeltaTableError`] if `data_folder_path` is not accessible.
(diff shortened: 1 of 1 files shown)


crates/modelardb_common/src/storage.rs
```diff
@@ -68,6 +69,10 @@ impl DeltaLake {
     /// Create a new [`DeltaLake`] that manages the delta tables in `data_folder_path` if it exists.
```
CGodiksen, line 69: Maybe update this doc comment now that it does not matter if it exists.

crates/modelardb_common/src/storage.rs
```diff
@@ -68,6 +69,10 @@ impl DeltaLake {
...
     /// Returns a [`DeltaTableError`] if `data_folder_path` is not accessible.
```
CGodiksen, line 70: Maybe change this to say that an error is returned if the path cannot be created if it does not already exist.  I am not sure `LocalFileSystem::new_with_prefix` can fail now that we create the path before.

CGodiksen (changes requested):

CGodiksen (approved):

chrthomsen (approved):

Issue #218, opened 2024-09-27
It is possible to register a node with the same URL twice in the manager metadata Delta Lake


CGodiksen:
When registering a node with a URL that is already registered, the manager should return an error. Currently an error is returned but before the error is thrown the node is saved to the metadata Delta Lake. When restarting the manager, the cluster is initialized with the duplicate nodes and the manager crashes.

The following steps can be taken to reproduce the bug:

1. Start the manager.
2. Register a node with the URL `grpc://127.0.0.1:9999`.
3. Register another node with the URL `grpc://127.0.0.1:9999`.
4. Stop the manager and start it again.
5. The manager is now not able to start since there are duplicate nodes in the metadata Delta Lake.

PR #230, opened 2024-09-30
No longer possible to register a node with the same URL twice in the manager metadata Delta Lake

CGodiksen:
This PR fixes https://github.com/ModelarData/ModelarDB-RS/issues/218 by reordering the actions done when registering a node with the manager. Now we save the node in the cluster first, which checks for duplicates, and then we save it in the metadata Delta Lake after.


crates/modelardb_manager/src/remote.rs
```diff
@@ -447,21 +447,23 @@ impl FlightService for FlightServiceHandler {
...
-            // and metadata Delta Lake will be out of sync until the manager is restarted.
+            // Use the metadata manager to persist the node to the metadata Delta Lake. Note that if
+            // this fails, the metadata Delta Lake and the cluster will be out of sync until the
+            // manager is restarted.
             self.context
-                .cluster
-                .write()
+                .remote_data_folder
+                .metadata_manager
+                .save_node(&node)
```
skejserjensen, line 465: Maybe change `save_node()` to take  ownership of `node` as it currently takes a reference but clones all members of `node`. By taking ownership of `node` it can reuse the members without of copying them and if the user only have a reference to `node` they can then copy instead of `save_node()` always making a copy.

crates/modelardb_manager/src/remote.rs
```diff
@@ -447,21 +447,23 @@ impl FlightService for FlightServiceHandler {
...
-                .save_node(&node)
+                .cluster
+                .write()
                 .await
+                .register_node(node.clone())
                 .map_err(|error| Status::internal(error.to_string()))?;
 
-            // Use the cluster to register the node in memory. Note that if this fails, the cluster
-            // and metadata Delta Lake will be out of sync until the manager is restarted.
+            // Use the metadata manager to persist the node to the metadata Delta Lake. Note that if
```
skejserjensen, line 459: Could the problem of the metadata Delta Lake and the manager being out of sync be remedied if the code checked if `save_node()` returned an error and deleted `node` from `cluster` before returning the error? Of course that leaves a small window where `node` could be used by `modelardbm`, but it seems better than leaving it registered. For a completely safe implementation a transactional system with a staging area is probably needed so changes can be staged until all involved have verified that they can perform the required task before all committing, but this seems out of scope for this PR as this should probably be done in a separate component used throughout the system due to how crucial it is to implement correctly and how easy it is to make mistakes.

skejserjensen (changes requested):

crates/modelardb_manager/src/remote.rs
CGodiksen: Fixed.

skejserjensen (approved):

chrthomsen (approved):

Issue #220, opened 2024-09-27
Server crashes when trying to ingest data when running with a manager


CGodiksen:
When starting a cluster deployment with a manager and a single edge, the edge crashes after ingesting data.

The following error message is returned:
`Failed to receive uncompressed message due to: Generic DeltaTable error: Error during planning: No table named 'test_model_table_1_tags'`

The table does exist in the metadata Delta Lake. It should be noted that when running a single edge deployment, the edge can ingest data, flush the memory, and query without problems.

Steps to reproduce:
1. Start the manager with MinIO as the remote data folder.
2. Start an edge node that is connected to the already started manager.
3. Create a model table named `test_model_table_1` on the manager which also creates it on the edge.
4. Ingest any amount of data into the table on the edge.
5. The edge should now crash or halt.

Issue #227, opened 2024-09-29
Use a single connection for data and metadata management


skejserjensen:
As management of [data](https://github.com/ModelarData/ModelarDB-RS/pull/185) and [metadata](https://github.com/ModelarData/ModelarDB-RS/pull/208) using [`deltalake`](https://crates.io/crates/deltalake) was developed in parallel, is code and configuration for connecting to, reading from, and writing to object stores of different types duplicated across the [`modelardb_common/src/storage.rs`](https://github.com/ModelarData/ModelarDB-RS/blob/main/crates/modelardb_common/src/storage.rs) and  [`modelardb_common/src/metadata`](https://github.com/ModelarData/ModelarDB-RS/tree/main/crates/modelardb_common/src/metadata) modules. This seems unnecessary and a single type for connecting to and performing operations on the object store in ModelarDB should be created.

CGodiksen:
When combining these into a single type we should also reconsider whether we want to add unit tests for the `DeltaLake` struct directly or continue to rely on other components testing the functionality indirectly.

PR #231, opened 2024-09-30
Fixed server crashing when trying to ingest data when running with a manager

CGodiksen:
This PR fixes https://github.com/ModelarData/ModelarDB-RS/issues/220 which was caused by the model table tag table not being registered in the table metadata manager for the remote data folder. This caused the server to crash when trying to save tag hash metadata to the remote data folder during ingestion when running with a manager.

The bug was fixed by using `open_table_with_storage_options()` when retrieving the schema of the table in the `MetadataDeltaLake` struct. Note that this is a very basic solution that should be fine since `MetadataDeltaLake` will be replaced when https://github.com/ModelarData/ModelarDB-RS/issues/227 is implemented.


crates/modelardb_common/src/metadata/mod.rs
```diff
@@ -194,10 +195,15 @@ impl MetadataDeltaLake {
...
     ) -> Result<DeltaTable, DeltaTableError> {
-        let table_provider = self.session.table_provider(table_name).await?;
-        let batch = RecordBatch::try_new(table_provider.schema(), rows)?;
+        let table = open_table_with_storage_options(
+            format!("{}/{table_name}", self.location),
+            self.storage_options.clone(),
+        )
+        .await?;
+
+        let batch = RecordBatch::try_new(TableProvider::schema(&table), rows)?;
```
skejserjensen, line 204: Maybe add a comment why `TableProvider::schema(&table)` is used instead of `table.schema()`, I assume it is so [`TableProvider::schema()`](https://docs.rs/deltalake/latest/deltalake/datafusion/catalog/trait.TableProvider.html#tymethod.schema) is called instead of [`DeltaTable.schema()`](https://docs.rs/deltalake/latest/deltalake/struct.DeltaTable.html#method.schema).

crates/modelardb_common/src/metadata/mod.rs
```diff
@@ -209,8 +215,13 @@ impl MetadataDeltaLake {
...
     ) -> Result<DataFrame, DeltaTableError> {
-        let table_provider = self.session.table_provider(table_name).await?;
-        let batch = RecordBatch::try_new(table_provider.schema(), rows)?;
+        let table = open_table_with_storage_options(
+            format!("{}/{table_name}", self.location),
+            self.storage_options.clone(),
+        )
+        .await?;
+
+        let batch = RecordBatch::try_new(TableProvider::schema(&table), rows)?;
```
skejserjensen, line 224: Maybe add a comment why `TableProvider::schema(&table)` is used instead of `table.schema()`, I assume it is so [`TableProvider::schema()`](https://docs.rs/deltalake/latest/deltalake/datafusion/catalog/trait.TableProvider.html#tymethod.schema) is called instead of [`DeltaTable.schema()`](https://docs.rs/deltalake/latest/deltalake/struct.DeltaTable.html#method.schema).

skejserjensen (approved):

crates/modelardb_common/src/metadata/mod.rs
CGodiksen: Fixed.

chrthomsen (approved):

Issue #222, opened 2024-09-28
Server crashes when trying to create table on a cloud node


CGodiksen:
When running a cluster deployment with a cloud node, the server crashes when a table is created on the manager and the manager tries to create the table on the cloud node.

The following error is returned when trying to create a table:
`Cluster Error: status: Internal, message: "Configuration Error: Not a Delta table: no log files"`

Steps to reproduce:

1. Start a manager with MinIO as the remote data folder.
2. Start a cloud node that is connected to the already started manager.
3. Create a model table named `test_model_table_1` on the manager which also creates it on the cloud.
4. The cloud node should now crash.

PR #232, opened 2024-09-30
Fixed server crashing when trying to create table on a cloud node

CGodiksen:
This PR fixes https://github.com/ModelarData/ModelarDB-RS/issues/222 which was caused by the cloud node trying to register a Delta table in the query data folder that did not exist yet.

The bug was fixed by the manager creating an empty Delta Lake table in the remote data folder when a table is created before the request is sent along to the edge and cloud nodes. This means that when the create table request gets to the cloud node, the table in the remote data folder can be registered.

The PR also fixes some minor documentation issues and another small bug where we registered model tables always using the table metadata manager from the local data folder instead of the query data folder. This is fine on edge nodes but on cloud nodes this meant that the node would query data from the remote data folder and tag metadata from the local data folder.


crates/modelardb_server/src/context.rs
```diff
@@ -222,13 +222,15 @@ impl Context {
     /// DataFusion. If the model tables could not be retrieved from the metadata Delta Lake or a
     /// model table could not be registered, return [`ModelarDbError`].
     pub async fn register_model_tables(&self) -> Result<(), ModelarDbError> {
-        let table_metadata_manager = &self.data_folders.local_data_folder.table_metadata_manager;
-
-        let model_table_metadata = table_metadata_manager
+        let model_table_metadata = self
```
skejserjensen, line 225: Maybe add a comment why `model_table_metadata` uses the `local_data_folder` while `table_metadata_manager` uses the `query_data_folder` as it is not clear.

skejserjensen (changes requested):

crates/modelardb_server/src/context.rs
CGodiksen: Added a comment that specifies that we use the local data folder since we don't want to register tables that the data sinks cannot write data to (since it uses the storage engine). The query data folder might contain more tables than the local data folder on start up but that should not be a problem since we retrieve and create the potentially missing tables from the manager right after registering tables and model tables locally.

skejserjensen (approved):

chrthomsen (approved):

PR #237, opened 2024-10-03
GitHub Actions Object Stores

### LINKED ISSUE #226: Add MinIO and Azurite to the GitHub Action Runners
Tests are currently only executed on local data as an object store is not available in [GitHub Action's Runners](https://github.com/actions/runner-images/tree/main). Thus, testing deployments with multiple instances of `modelardbd` across edge and cloud an an instance of `modelardm` is not done. To remedy this, both [MinIO](https://min.io/) and [Azurite](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite) should be added to the GitHub Action Runners so tests with all possible configurations can be performed. Both MinIO and Azurite is needed, as ModelarDB is designed to support multiple different object stores and both S3 and Azure support is currently implemented with S3 being the most mature. Possible solutions for  installing and running MinIO and Azurite in the background across all three runners could be:

- **Docker:** Simple to deploy, but this is not possible on macOS as it is only installed on Linux and Windows.
- **Perl/Python/Ruby/etc Script:** It is unclear how to run the script in the background across all three OSes.
- **GitHub Actions Marketplace:** [minio-action](https://github.com/infleet/minio-action) assumes sh and Docker, while [azuright](https://github.com/potatoqualitee/azuright) assumes PowerShell.
- **One yml File with per OS ifs:** A more complex GitHub Action yml file seems costly as it is hard to debug.
- **Separate yml Files per OS:** Multiple files to updated, but < 100 lines, and simple to run OS commands.

skejserjensen:
This PR fixes #226 by adding Minio and Azurite to the GitHub Actions runners for use with testing. A separate `yml` file was not created for each platform was agreed in #226 as the only part of the `yml` file that needed to be specialized for each platform was specifying the version of Minio to download. As `deltalake` and `object_store` seems to lack functionality for creating S3 buckets and Azure Blob Storage containers, an S3 bucket named `modelardb` and an Azure Blob Storage container named `modelardb` is created so tests that use the object stores can be implemented without adding additional dependencies.


.github/workflows/cargo-build-lint-and-test.yml
```diff
@@ -21,11 +21,19 @@ on:
     branches: [ "main" ]
 
 env:
+  # General.
   CARGO_TERM_COLOR: always
   RUSTFLAGS: -C instrument-coverage -C strip=symbols -D warnings
   RUSTDOCFLAGS: -D warnings
   LLVM_PROFILE_FILE: modelardb-%p-%m.profraw
 
+  # Minio.
```
CGodiksen, line 30:
```suggestion
  # MinIO.
```

.github/workflows/cargo-build-lint-and-test.yml
```diff
@@ -21,11 +21,19 @@ on:
...
+  AWS_ACCESS_KEY_ID: minioadmin
+  AWS_SECRET_ACCESS_KEY: minioadmin
+  AWS_DEFAULT_REGION: us-west-1
```
CGodiksen, line 33: I don't think it does anything but maybe change to `eu-central-1` for consistency with the Docker Compose file.

.github/workflows/cargo-build-lint-and-test.yml
```diff
@@ -36,9 +44,36 @@ jobs:
         operating-system: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
+    # Retrieve Code.
     - name: Checkout
       uses: actions/checkout@v4
 
+    # Setup Minio.
```
CGodiksen, line 51:
```suggestion
    # Setup MinIO.
```

.github/workflows/cargo-build-lint-and-test.yml
```diff
@@ -36,9 +44,36 @@ jobs:
...
+    - name: Setup Minio
```
CGodiksen, line 52:
```suggestion
    - name: Setup MinIO
```

CGodiksen (approved):

.github/workflows/cargo-build-lint-and-test.yml
skejserjensen: Fixed.

.github/workflows/cargo-build-lint-and-test.yml
skejserjensen: Fixed.

.github/workflows/cargo-build-lint-and-test.yml
skejserjensen: Fixed.

.github/workflows/cargo-build-lint-and-test.yml
skejserjensen: Fixed.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -55,7 +55,8 @@ const HOST: &str = "127.0.0.1";
 
 /// The next port to be used for the server in an integration test. Each test uses a unique port and
 /// local data folder, so they can run in parallel and so that any failing tests do not cascade.
-static PORT: AtomicU16 = AtomicU16::new(9999);
+/// Ports 9000 and 9001 is used by MinIO and Ports 10000, 10001, and 10002 is used by Azurite.
```
chrthomsen, line 58: 2x "is" -> "are"

chrthomsen (approved):

Issue #246, opened 2024-10-11
Use best practices for Errors and Panics


skejserjensen:
Currently the ModelarDB server indicates errors by returning error types defined by our dependencies, returning error types we define, returning strings, and a few `panic!()` to indicate that an error have occurred. However, at the moment there is no consistency when which method is used, so this task is to determine how the Rust community generally recommends errors should be managed so we can follow best practices, e.g, using resources like [this](https://nrc.github.io/error-docs/). 

Also, it would be beneficial if the use of these best practices could remove all uses of `panic!()` so the system does not simply crash when an error occurs, e.g., rejecting new data points by returning an error if the storage engine is out of memory for new data points instead of simply calling `panic!()`. In the best case the code should not contain panic!() or use API's that may `panic!()` to prevent the server from crashing and stopping ingestion when an error occurs.

PR #247, opened 2024-10-11
Split common into common and types

skejserjensen:
This PR splits `modelardb_common` into `modelardb_common` and `modelardb_types` so `ModelarDbError` can be used without a lot of dependencies in preparation for working on #246.


crates/modelardb_types/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+# Copyright 2023 The ModelarDB Contributors
```
chrthomsen, line 1: 2024 ?

crates/modelardb_types/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+/* Copyright 2023 The ModelarDB Contributors
```
chrthomsen, line 1: 2024 ?

chrthomsen (approved):

crates/modelardb_types/Cargo.toml
skejserjensen: Fixed.

crates/modelardb_types/src/lib.rs
skejserjensen: Fixed.

CGodiksen (approved):

Issue #268, opened 2024-12-09
It is possible to create and drop tables directly on a server when running in a cluster


CGodiksen:
The operations were moved to `do_get` in https://github.com/ModelarData/ModelarDB-RS/pull/266 and we relied on `validate_action_request` to stop creating and dropping tables directly on a server node when running in a cluster. Currently it is possible to create or drop a table directly on a server which would result in an inconsistent database schema across the cluster.

Issue #267, opened 2024-12-09
Manager still uses actions for creating, dropping, and truncating tables for each node in a cluster


CGodiksen:
Since creating, dropping, and truncating tables have been moved to `do_get()` in https://github.com/ModelarData/ModelarDB-RS/pull/266, the process for doing these actions for each node in a cluster should also be updated. Currently the manager still uses the old actions which result in an `Action not implemented` error if trying to, for example, create a table in a cluster.

PR #269, opened 2024-12-10
Fixed bugs caused by moving actions to do_get

CGodiksen:
This PR fixes bugs that were caused by moving certain operations to `do_get` from `do_action` in https://github.com/ModelarData/ModelarDB-RS/pull/266. Specifically, the PR fixes https://github.com/ModelarData/ModelarDB-RS/issues/268 which was a bug that allowed creating and dropping tables directly on a server even when running in a cluster and https://github.com/ModelarData/ModelarDB-RS/issues/267 which was a bug caused by the `Cluster` type in the manager still using the old actions to create, drop, and truncate.

The PR also fixes another minor bug that was caused by the return value in the `do_get` method in the manager. The empty stream cased a `Server never sent a data message` error on the calling client which has been fixed by returning flight data with an empty schema in it.


crates/modelardb_manager/src/cluster.rs
```diff
@@ -137,84 +137,42 @@ impl Cluster {
...
         }
 
         Ok(())
     }
 
-    /// Connect to the Apache Arrow flight client given by `url` and make a request to do `action`.
-    /// If the action was successfully executed, return the url of the node, otherwise return
-    /// [`ModelarDbManagerError`].
-    async fn connect_and_do_action(
+    /// Connect to the Apache Arrow flight client given by `url` and execute the given `sql`
```
skejserjensen, line 160: Is the flight client not used to connect to the instance of `modelardbd` at `url` instead of a Apache Arrow flight client? Maybe rephrase the comment.

crates/modelardb_manager/src/cluster.rs
```diff
@@ -137,84 +137,42 @@ impl Cluster {
...
+    /// statement with the given `key`. If the statement was successfully executed, return the url
```
skejserjensen, line 161: Maybe "... with the given `key` as metadata." to make it more clear how `key` is used.

crates/modelardb_manager/src/remote.rs
```diff
@@ -460,7 +460,11 @@ impl FlightService for FlightServiceHandler {
         };
 
         // Confirm the SQL statement was executed.
-        Ok(Response::new(Box::pin(stream::empty())))
+        let options = IpcWriteOptions::default();
```
skejserjensen, line 463: Unclear why [`stream::empty()` ](https://docs.rs/futures/latest/futures/prelude/stream/fn.empty.html) was replaced with code that also seems to create an empty stream. Maybe extend the comment to state what the stream returns, and if it is empty, why `stream::empty()` is not used.

crates/modelardb_server/src/remote.rs
```diff
@@ -427,6 +443,7 @@ impl FlightService for FlightServiceHandler {
         &self,
         request: Request<Ticket>,
     ) -> StdResult<Response<Self::DoGetStream>, Status> {
+        let request_metadata = request.metadata().clone();
```
skejserjensen, line 446: Unclear why the `MetadataMap` is cloned when the type of `validate_request`'s argument is `&MetadataMap` which [`Request.metadata()`](https://docs.rs/tonic/latest/tonic/struct.Request.html#method.metadata) return.

crates/modelardb_server/src/remote.rs
```diff
@@ -602,21 +625,10 @@ impl FlightService for FlightServiceHandler {
         &self,
         request: Request<Action>,
     ) -> StdResult<Response<Self::DoActionStream>, Status> {
-        let metadata = request.metadata().clone();
+        let request_metadata = request.metadata().clone();
```
skejserjensen, line 628: Unclear why the `MetadataMap` is cloned when the type of `validate_request`'s argument is `&MetadataMap` which [`Request.metadata()`](https://docs.rs/tonic/latest/tonic/struct.Request.html#method.metadata) return.

skejserjensen (changes requested):

crates/modelardb_manager/src/cluster.rs
CGodiksen: Fixed.

crates/modelardb_manager/src/remote.rs
CGodiksen: The new code does not create an empty stream, it just creates a stream of a single element. It seems that a schema is required for `do_get` to execute correctly in PyArrow and possibly also other clients. I have made it clear in the comment that the stream is not empty.

crates/modelardb_server/src/remote.rs
CGodiksen: It was required because of the call to `request.into_inner()` on the next line but changing it to `request.get_ref()` fixed the issue.

skejserjensen (approved):

chrthomsen (approved):

Issue #270, opened 2024-12-15
Cloud nodes return an error when trying to query model tables created through the manager


CGodiksen:
Currently cloud nodes return an error if you query a table that was created through the manager while the cloud node is running. This is caused by the `model_table_name_tags` table not being registered with the query data folder table metadata manager session context. This is not a problem for edge nodes since the query data folder is the same as the local data folder.

PR #271, opened 2024-12-16
Fixed bug causing cloud nodes to return error when querying

CGodiksen:
This PR fixes https://github.com/ModelarData/ModelarDB-RS/issues/270 which caused cloud nodes to return an error when trying to query a table that was created while the cloud node was running.

The PR adds a new method to the table metadata manager that can be used to register the `model_table_name_tags` table which is the only metadata table required during querying. This new method is used to register the metadata table in the query data folder table metadata manager when a table is created.

The PR also adds unit tests for this new method since it is public.

### DIFF
--- a/crates/modelardb_server/src/context.rs
+++ b/crates/modelardb_server/src/context.rs
@@ -146,7 +146,7 @@ impl Context {
             .create_model_table(&model_table_metadata.name)
             .await?;
 
-        let query_table_metadata_manager = self
+        let query_folder_table_metadata_manager = self
             .data_folders
             .query_data_folder
             .table_metadata_manager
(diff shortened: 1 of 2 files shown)


crates/modelardb_server/src/context.rs
```diff
@@ -146,23 +146,33 @@ impl Context {
             .create_model_table(&model_table_metadata.name)
             .await?;
 
-        let table_metadata_manager = self
+        let query_table_metadata_manager = self
```
skejserjensen, line 149: Maybe `query_folder_table_metadata_manager` as I did not understand what `quey_` was until I read how the variable was created.

skejserjensen (approved):

crates/modelardb_server/src/context.rs
CGodiksen: Fixed.

Issue #272, opened 2024-12-17
Add support for creating tables using schema, error bounds, and generated columns directly


CGodiksen:
Currently we parse SQL `CREATE TABLE` statements into a table name, schema, possible error bounds, and possible generated columns to create normal tables and model tables. However, in some cases it can be ideal to support creating tables directly using this table data instead of having to parse the table data into a SQL statement and then parse it back in ModelarDB. 

Furthermore, if it is possible to create tables using table data directly, then this can also be used when we initialize the database of a server node in a cluster to avoid having to save the SQL that was used to create a table. This could simplify the metadata Delta Lake further.

PR #275, opened 2025-01-19
Added new CreateTables action to create tables using table metadata

CGodiksen:
This PR implements https://github.com/ModelarData/ModelarDB-RS/issues/272 by adding a new `CreateTables` action to the server and manager that can be used to create tables using table metadata instead of a `CREATE TABLE` SQL statement.

This new action takes a record batch encoded as bytes in the action body and creates the tables with the type, name, schema, potential error bounds, and potential generated columns in the record batch. Note that multiple tables can be created with a single call to `CreateTables`.

Since we can now create tables without SQL, the initialization process for server nodes in a cluster has been changed to no longer use SQL to initialize the database schema of the added node. Instead, the table metadata for the missing tables are transferred from the manager to the new server node, and the same process as in the new action is used to create the tables.

This means that it was no longer necessary to save `CREATE TABLE` SQL statements in the metadata delta lake so those columns have been removed as well as the metadata delta lake functionality related to the SQL columns.

### DIFF
--- a/crates/modelardb_manager/src/cluster.rs
+++ b/crates/modelardb_manager/src/cluster.rs
@@ -17,6 +17,7 @@
 
 use std::collections::VecDeque;
 
+use arrow::record_batch::RecordBatch;
 use arrow_flight::flight_service_client::FlightServiceClient;
 use arrow_flight::{Action, Ticket};
 use futures::stream::FuturesUnordered;
@@ -135,6 +136,22 @@ impl Cluster {
                 "There are no cloud nodes to execute the query in the cluster.".to_owned(),
             ))
         }
+    }
+
+    /// Create tables on each node in the cluster with the given `record_batch` and `key` as metadata.
+    /// If the tables were successfully created on each node, return [`Ok`], otherwise return
+    /// [`ModelarDbManagerError`].
+    pub async fn create_tables(
+        &self,
+        record_batch: &RecordBatch,
+        key: &MetadataValue<Ascii>,
+    ) -> Result<()> {
+        let action = Action {
+            r#type: "CreateTables".to_owned(),
+            body: modelardb_storage::try_convert_record_batch_to_bytes(record_batch)?.into(),
+        };
+
+        self.cluster_do_action(action, key).await
     }
 
     /// For each node in the cluster, execute the given `sql` statement with the given `key` as
--- a/crates/modelardb_manager/src/cluster.rs
+++ b/crates/modelardb_manager/src/cluster.rs
@@ -176,6 +176,51 @@ impl Cluster {
 
         Ok(url.to_owned())
     }
+
+    /// For each node in the cluster, execute the given `action` with the given `key` as metadata.
+    /// If the action was successfully executed for each node, return [`Ok`], otherwise return
+    /// [`ModelarDbManagerError`].
+    pub async fn cluster_do_action(
+        &self,
+        action: Action,
+        key: &MetadataValue<Ascii>,
+    ) -> Result<()> {
+        let mut action_futures: FuturesUnordered<_> = self
+            .nodes
+            .iter()
+            .map(|node| self.connect_and_do_action(&node.url, action.clone(), key))
+            .collect();
+
+        // Run the futures concurrently and log when the action has been executed on each node.
+        while let Some(result) = action_futures.next().await {
+            info!(
+                "Executed action `{}` on node with url '{}'.",
+                action.r#type, result?
+            );
+        }
+
+        Ok(())
+    }
+
+    /// Connect to the Apache Arrow flight server given by `url` and make a request to do `action`
+    /// with the given `key` as metadata. If the action was successfully executed, return the url
+    /// of the node, otherwise return [`ModelarDbManagerError`].
+    async fn connect_and_do_action(
+        &self,
+        url: &str,
+        action: Action,
+        key: &MetadataValue<Ascii>,
+    ) -> Result<String> {
+        let mut flight_client = FlightServiceClient::connect(url.to_owned()).await?;
+
+        // Add the key to the request metadata to indicate that the request is from the manager.
+        let mut request = Request::new(action);
+        request.metadata_mut().insert("x-manager-key", key.clone());
+
+        flight_client.do_action(request).await?;
+
+        Ok(url.to_owned())
+    }
 }
 
 impl Default for Cluster {
--- a/crates/modelardb_manager/src/remote.rs
+++ b/crates/modelardb_manager/src/remote.rs
@@ -25,7 +25,6 @@ use std::str::FromStr;
 use std::sync::Arc;
 
 use arrow::datatypes::Schema;
-use arrow::error::ArrowError;
 use arrow::ipc::reader::StreamReader;
 use arrow::ipc::writer::IpcWriteOptions;
 use arrow_flight::flight_service_server::{FlightService, FlightServiceServer};
@@ -38,6 +37,7 @@ use bytes::Buf;
 use futures::{stream, Stream};
 use modelardb_common::arguments;
 use modelardb_common::remote;
+use modelardb_common::remote::{error_to_status_internal, error_to_status_invalid_argument};
 use modelardb_storage::metadata::model_table_metadata::ModelTableMetadata;
 use modelardb_storage::parser;
 use modelardb_storage::parser::ModelarDbStatement;
@@ -105,7 +105,7 @@ impl FlightServiceHandler {
         let schema = if table_metadata_manager
             .is_normal_table(&table_name)
             .await
-            .map_err(|error| Status::internal(error.to_string()))?
+            .map_err(error_to_status_internal)?
         {
             let delta_table = self
                 .context
@@ -113,25 +113,24 @@ impl FlightServiceHandler {
                 .delta_lake
                 .delta_table(&table_name)
                 .await
-                .map_err(|error| Status::internal(error.to_string()))?;
+                .map_err(error_to_status_internal)?;
 
-            // unwrap() is safe since the schema is always valid.
             let schema = delta_table
                 .get_schema()
-                .map_err(|error| Status::internal(error.to_string()))?
+                .map_err(error_to_status_internal)?
                 .try_into()
-                .unwrap();
+                .map_err(error_to_status_internal)?;
 
             Arc::new(schema)
         } else if table_metadata_manager
             .is_model_table(&table_name)
             .await
-            .map_err(|error| Status::internal(error.to_string()))?
+            .map_err(error_to_status_internal)?
         {
             let model_table_metadata = table_metadata_manager
                 .model_table_metadata_for_model_table(&table_name)
                 .await
-                .map_err(|error| Status::internal(error.to_string()))?;
+                .map_err(error_to_status_internal)?;
 
             model_table_metadata.query_schema
         } else {
@@ -152,7 +151,7 @@ impl FlightServiceHandler {
             .metadata_manager
             .table_metadata_column("table_name")
             .await
-            .map_err(|error| Status::internal(error.to_string()))?;
+            .map_err(error_to_status_internal)?;
 
         if existing_tables
             .iter()
@@ -180,7 +179,7 @@ impl FlightServiceHandler {
             .delta_lake
             .create_normal_table(table_name, schema)
             .await
-            .map_err(|error| Status::internal(error.to_string()))?;
+            .map_err(error_to_status_internal)?;
 
         // Persist the new normal table to the metadata Delta Lake.
         self.context
@@ -189,7 +188,7 @@ impl FlightServiceHandler {
             .table_metadata_manager
             .save_normal_table_metadata(table_name, sql)
             .await
-            .map_err(|error| Status::internal(error.to_string()))?;
+            .map_err(error_to_status_internal)?;
 
         // Register and save the table to each node in the cluster.
         self.context
@@ -198,7 +197,7 @@ impl FlightServiceHandler {
             .await
             .cluster_do_get(sql, &self.context.key)
             .await
-            .map_err(|error| Status::internal(error.to_string()))?;
+            .map_err(error_to_status_internal)?;
 
         info!("Created normal table '{}'.", table_name);
 
@@ -219,7 +218,7 @@ impl FlightServiceHandler {
             .delta_lake
             .create_model_table(&model_table_metadata.name)
             .await
-            .map_err(|error| Status::internal(error.to_string()))?;
+            .map_err(error_to_status_internal)?;
 
         // Persist the new model table to the metadata Delta Lake.
         self.context
@@ -228,7 +227,7 @@ impl FlightServiceHandler {
             .table_metadata_manager
             .save_model_table_metadata(&model_table_metadata, sql)
             .await
-            .map_err(|error| Status::internal(error.to_string()))?;
+            .map_err(error_to_status_internal)?;
 
         // Register and save the model table to each node in the cluster.
         self.context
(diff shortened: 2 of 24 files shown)


crates/modelardb_storage/src/lib.rs
```diff
@@ -311,13 +321,264 @@ fn apache_parquet_writer_properties(
...
+/// Convert a list of optional [`GeneratedColumns`](GeneratedColumn) to a [`ListArray`].
+fn generated_columns_to_list_array(generated_columns: Vec<Option<GeneratedColumn>>) -> ListArray {
+    let mut generated_columns_builder = ListBuilder::new(StringBuilder::new());
+
+    for generated_column in generated_columns {
+        if let Some(generated_column) = generated_column {
+            if let Some(sql_expr) = generated_column.original_expr {
+                generated_columns_builder.values().append_value(sql_expr);
+            } else {
+                generated_columns_builder.values().append_null();
```
skejserjensen, line 460: If a generated column exists for one table in the cluster will it not created issues if it does not exist for other instances of the table on other nodes in the same cluster? Seems generated columns for which the original expression cannot be extracted should be rejected when the table is created as they cannot be propagated to other nodes in the cluster.

crates/modelardb_storage/src/lib.rs
```diff
@@ -311,13 +321,264 @@ fn apache_parquet_writer_properties(
...
+                    Arc::new(schema),
+                    error_bounds,
+                    generated_columns,
+                )?;
+
+                model_table_metadata.push(metadata);
+            }
+            _ => {
+                return Err(ModelarDbStorageError::InvalidArgument(format!(
+                    "Table type '{table_type}' is not supported.",
```
skejserjensen, line 520: Should a boolean array be used for table type to make it impossible to represent anything but normal tables and model tables?

crates/modelardb_storage/src/lib.rs
```diff
@@ -311,13 +321,264 @@ fn apache_parquet_writer_properties(
...
+    }
+
+    Ok((normal_table_metadata, model_table_metadata))
+}
+
+/// Parse the error bound values in `error_bounds_array` into a list of [`ErrorBounds`](ErrorBound).
+/// Returns [`ModelarDbServerError`] if an error bound value is invalid.
+fn array_to_error_bounds(error_bounds_array: ArrayRef) -> Result<Vec<ErrorBound>> {
+    // unwrap() is safe since error bound values are always f32.
+    let value_array = error_bounds_array
```
skejserjensen, line 533: Maybe use the `array!()` macro to shorten the code and make it easy to return the error in the future for casts by changing the macro.

crates/modelardb_storage/src/lib.rs
```diff
@@ -311,13 +321,264 @@ fn apache_parquet_writer_properties(
...
+
+/// Parse the generated column expressions in `generated_columns_array` into a list of optional
+/// [`GeneratedColumns`](GeneratedColumn). Returns [`ModelarDbServerError`] if a generated column
+/// expression is invalid.
+fn array_to_generated_columns(
+    generated_columns_array: ArrayRef,
+    df_schema: &DFSchema,
+) -> Result<Vec<Option<GeneratedColumn>>> {
+    // unwrap() is safe since generated column expressions are always strings.
+    let expr_array = generated_columns_array
```
skejserjensen, line 558: Maybe use the `array!()` macro to shorten the code and make it easy to return the error in the future for casts by changing the macro.

crates/modelardb_storage/src/lib.rs
```diff
@@ -446,4 +707,172 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_invalid_bytes_to_schema() {
+        assert!(try_convert_bytes_to_schema(vec!(1, 2, 4, 8)).is_err());
+    }
+
+    // Tests for try_convert_record_batch_to_bytes() and try_convert_bytes_to_record_batch().
+    #[test]
+    fn test_convert_record_batch_to_bytes_and_bytes_to_record_batch() {
```
skejserjensen, line 734: Maybe add a test with a model table to ensure it works for both table types.

crates/modelardb_storage/src/lib.rs
```diff
@@ -446,4 +707,172 @@ mod tests {
...
+
+        let expected_schema_bytes =
+            try_convert_schema_to_bytes(&model_table_metadata.query_schema).unwrap();
+        assert_eq!(
+            **record_batch.column(2),
+            BinaryArray::from_vec(vec![&expected_schema_bytes])
+        );
+
+        let error_bounds_array = modelardb_types::array!(record_batch, 3, ListArray).value(0);
+        let value_array = error_bounds_array
```
skejserjensen, line 812: Maybe use the `array!()` macro to shorten the code and make it easy to return the error in the future for casts by changing the macro.

crates/modelardb_storage/src/lib.rs
```diff
@@ -446,4 +707,172 @@ mod tests {
...
+            .as_any()
+            .downcast_ref::<Float32Array>()
+            .unwrap();
+
+        assert_eq!(value_array, &Float32Array::from(vec![0.0, 1.0, -5.0, -0.0]));
+
+        let generated_columns_array = modelardb_types::array!(record_batch, 4, ListArray).value(0);
+        let expr_array = generated_columns_array
```
skejserjensen, line 820: Maybe use the `array!()` macro to shorten the code and make it easy to return the error in the future for casts by changing the macro.

crates/modelardb_storage/src/metadata/table_metadata_manager.rs
```diff
@@ -1103,6 +1046,23 @@ mod tests {
             .unwrap());
     }
```
skejserjensen, line 1048: Maybe add a comment documenting which functions are tested as is done elsewhere.

crates/modelardb_manager/src/cluster.rs
```diff
@@ -137,6 +138,22 @@ impl Cluster {
         }
     }
 
+    /// Create tables on each node in the cluster with the given `record_batch` and `key` as metadata.
+    /// If the tables were successfully created on each node, return [`Ok`], otherwise return
+    /// [`ModelarDbManagerError`].
+    pub async fn create_tables(
+        &self,
+        record_batch: &RecordBatch,
```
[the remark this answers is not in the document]
CGodiksen: Fixed by now checking the schema of the record batch before executing the action.

crates/modelardb_manager/src/cluster.rs
```diff
@@ -176,6 +193,51 @@ impl Cluster {
...
+            info!(
+                "Executed action `{}` on node with url '{}'.",
+                action.r#type, result?
+            );
+        }
+
+        Ok(())
+    }
+
+    /// Connect to the Apache Arrow flight server given by `url` and make a request to do `action`
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

crates/modelardb_manager/src/cluster.rs
```diff
@@ -176,6 +193,51 @@ impl Cluster {
...
+    /// with the given `key` as metadata. If the action was successfully executed, return the url
```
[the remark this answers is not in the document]
CGodiksen: Added text that mentions that the URL is returned to simplify logging.

crates/modelardb_manager/src/remote.rs
```diff
@@ -529,31 +588,52 @@ impl FlightService for FlightServiceHandler {
...
+
+                        modelardb_storage::model_table_metadata_record_batch(&model_table_metadata)
+                            .map_err(error_to_status_internal)?
+                    };
+
+                    record_batches.push(record_batch);
                 }
 
-                let response_body = table_sql_queries.join(";").as_bytes().to_vec();
+                let record_batch =
```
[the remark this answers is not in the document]
CGodiksen: I agree that it would probably be slightly more efficient but I have kept it as it was since the code is not run often and I prefer using a cleaner interface that can also be used elsewhere.

crates/modelardb_server/src/context.rs
```diff
@@ -70,17 +71,34 @@ impl Context {
...
-    pub(crate) async fn create_normal_table(
-        &self,
-        name: String,
-        schema: Schema,
-        sql: &str,
-    ) -> Result<()> {
-        self.check_if_table_exists(&name).await?;
-        self.register_and_save_normal_table(&name, sql, schema)
-            .await?;
+    /// Convert the bytes to a record batch and create tables using the metadata in the record
```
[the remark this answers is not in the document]
CGodiksen: Fixed, also fixed in the other doc comments where we wrote `record batch`.

crates/modelardb_server/src/context.rs
```diff
@@ -154,7 +169,7 @@ impl Context {
 
         // Register the model table with Apache DataFusion.
         self.register_model_table(
-            model_table_metadata.clone(),
+            Arc::new(model_table_metadata.clone()),
```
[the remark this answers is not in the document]
CGodiksen: I have kept it as it was since I think the current structure allows the most borrows and only creates the `Arc` with the clone when necessary.

crates/modelardb_server/src/context.rs
```diff
@@ -434,19 +449,44 @@ impl Context {
...
     use super::*;
 
-    use modelardb_storage::parser;
-    use modelardb_storage::parser::ModelarDbStatement;
     use modelardb_storage::test;
     use tempfile::TempDir;
 
     use crate::data_folders::DataFolder;
 
     #[tokio::test]
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

crates/modelardb_server/src/storage/uncompressed_data_buffer.rs
```diff
@@ -157,21 +157,18 @@ impl UncompressedInMemoryDataBuffer {
...
         let mut columns = Vec::with_capacity(1 + self.values.len());
-        columns.push(compute::take(&timestamps, &sorted_indices, None).unwrap());
+        columns.push(compute::take(&timestamps, &sorted_indices, None)?);
         for value in &mut self.values {
-            columns.push(compute::take(&value.finish(), &sorted_indices, None).unwrap());
+            columns.push(compute::take(&value.finish(), &sorted_indices, None)?);
         }
 
-        // unwrap() is safe as uncompressed_schema only contains timestamps and values.
         Ok(RecordBatch::try_new(
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -1311,3 +1312,52 @@ fn test_can_get_node_type() {
 
     assert_eq!(str::from_utf8(&response_bytes).unwrap(), "server");
 }
+
+#[test]
```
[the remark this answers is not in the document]
CGodiksen: It does not seem like we do this in the integration tests. Instead it seems that the module doc comment serves this function and I have therefore left it as it was.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -1311,3 +1312,52 @@ fn test_can_get_node_type() {
...
+        modelardb_storage::try_convert_record_batch_to_bytes(&table_record_batch).unwrap();
+
+    let action = Action {
+        r#type: "CreateTables".to_owned(),
+        body: table_record_batch_bytes.into(),
+    };
+
+    test_context
+        .runtime
+        .block_on(async { test_context.client.do_action(Request::new(action)).await })
```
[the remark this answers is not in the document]
CGodiksen: An [Issue](https://github.com/ModelarData/ModelarDB-RS/issues/277) has been created to do this in a separate PR.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -1311,3 +1312,52 @@ fn test_can_get_node_type() {
...
+        modelardb_storage::try_convert_record_batch_to_bytes(&invalid_record_batch).unwrap();
+
+    let action = Action {
+        r#type: "CreateTables".to_owned(),
+        body: invalid_record_batch_bytes.into(),
+    };
+
+    let response = test_context
+        .runtime
+        .block_on(async { test_context.client.do_action(Request::new(action)).await });
```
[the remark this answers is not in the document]
CGodiksen: An [Issue](https://github.com/ModelarData/ModelarDB-RS/issues/277) has been created to do this in a separate PR.

crates/modelardb_storage/src/lib.rs
```diff
@@ -311,13 +321,264 @@ fn apache_parquet_writer_properties(
...
+        record_batches.push(record_batch);
+    }
+
+    concat_batches(schema, &record_batches).map_err(|error| error.into())
+}
+
+/// Return a [`RecordBatch`] constructed from the metadata of a normal table with the name
+/// `table_name` and the schema `schema`. If the schema could not be converted to bytes or the
+/// record batch could not be created, return [`ModelarDbStorageError`].
+pub fn normal_table_metadata_record_batch(
```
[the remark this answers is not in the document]
CGodiksen: Fixed.

crates/modelardb_storage/src/lib.rs
CGodiksen: I am also not entirely sure why `original_expr` in `GeneratedColumn` is `Option<String>` and not just `String`. If we want to change this, I can do this in a separate PR since I do not think it is relevant to this PR.

crates/modelardb_storage/src/lib.rs
CGodiksen: `array!()` is used to extract a column from a record batch which is not the case here. Since this pattern is only used a few times in this file, I have not added a new macro for this specific case.

crates/modelardb_storage/src/lib.rs
CGodiksen: `array!()` is used to extract a column from a record batch which is not the case here. Since this pattern is only used a few times in this file, I have not added a new macro for this specific case.

crates/modelardb_storage/src/lib.rs
CGodiksen: `array!()` is used to extract a column from a record batch which is not the case here. Since this pattern is only used a few times in this file, I have not added a new macro for this specific case.

crates/modelardb_storage/src/lib.rs
CGodiksen: `array!()` is used to extract a column from a record batch which is not the case here. Since this pattern is only used a few times in this file, I have not added a new macro for this specific case.

crates/modelardb_storage/src/lib.rs
CGodiksen: Good idea, fixed.

CGodiksen: Good idea, fixed.

crates/modelardb_storage/src/lib.rs
skejserjensen: Yes let us make the change in a separate PR, just resolve the conversation when an issue has been created.

crates/modelardb_storage/src/lib.rs
skejserjensen: Ups, for some reason I thought of `error_bounds_array` as `RecordBatch`.

crates/modelardb_storage/src/lib.rs
skejserjensen: Ups, for some reason I thought of `error_bounds_array` as `RecordBatch`.

crates/modelardb_storage/src/lib.rs
skejserjensen: Ups, for some reason I thought of `error_bounds_array` as `RecordBatch`.

skejserjensen (approved):

chrthomsen (approved):

PR #282, opened 2025-02-02
Add cast!() macro

skejserjensen:
PR #275 showed that casts from `Any` to a concrete type is sometimes needed outside of extracting arrays from `Vec`s and `Recordbatch`es. Thus, this PR adds the `cast!()` macro which is defined as `$expr.as_any().downcast_ref::<$type>().unwrap()` and implements the `value!()` and `array!()` macros in terms of it. The PR also replaces all uses of `as_any().downcast_ref().unwrap()` found by `ripgrep`  with the macro.

### DIFF
--- a/crates/modelardb_server/src/storage/uncompressed_data_manager.rs
+++ b/crates/modelardb_server/src/storage/uncompressed_data_manager.rs
@@ -212,24 +212,21 @@ impl UncompressedDataManager {
 
         // Prepare the timestamp column for iteration.
         let timestamp_index = model_table_metadata.timestamp_column_index;
-        let timestamp_column_array: &TimestampArray = data_points
-            .column(timestamp_index)
-            .as_any()
-            .downcast_ref()
-            .unwrap();
+        let timestamp_column_array: &TimestampArray =
+            modelardb_types::array!(data_points, timestamp_index, TimestampArray);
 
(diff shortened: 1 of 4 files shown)


crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -212,24 +212,21 @@ impl UncompressedDataManager {
 
         // Prepare the timestamp column for iteration.
         let timestamp_index = model_table_metadata.timestamp_column_index;
-        let timestamp_column_array: &TimestampArray = data_points
-            .column(timestamp_index)
-            .as_any()
-            .downcast_ref()
-            .unwrap();
+        let timestamp_column_array: &TimestampArray =
```
CGodiksen, line 215: Maybe remove the type hint since it seems that we do not use it elsewhere when we use `array!`.

CGodiksen (approved):

crates/modelardb_server/src/storage/uncompressed_data_manager.rs
skejserjensen: I have removed most of the annotations, but two `Vec<_>` are needed so the compiler which data structures to collect arrays into.

chrthomsen (approved):

Issue #284, opened 2025-02-10
Redesign how we handle modelardbds metrics


CGodiksen:
Currently we track metrics such as used ingested memory, used uncompressed memory, used compressed memory, ingested data points, and used disk space by manually appending values to the different metrics when a relevant change happens. This both makes the code hard to read but is also error prone since it relies on always remembering to append the values to the metric whenever anything relevant happens. Furthermore, the metrics are not currently used outside ModelarDB, which makes the feature not currently worth the added complexity.

Metrics should be redesigned and reimplemented so that we instead use existing functionality for metrics, for example `DataFusion`, and rely on tracking metrics automatically instead. Automatic tracking of metrics could be achieved by integrating them into `memory_pool` in `StorageEngine` and the construction and dropping of buffers. Currently we also use an action to retrieve the metrics, this could be changed so we instead expose the metrics using `TableProvider`. 

This feature could be implemented together with https://github.com/ModelarData/ModelarDB-RS/issues/283 as part of a larger refactor of how we manage memory in `StorageEngine`.

- [x] Remove the existing implementation of metrics.
- [ ] Redesign how we handle metrics.
- [ ] Reimplement metrics according to the new design.

skejserjensen:
Maybe metrics could be stored in a model table in the [`Information Schema`](https://datafusion.apache.org/user-guide/sql/information_schema.html). By making it a model table, the metrics should be efficiently stored and queried, and by placing the table in the `Information Schema` it will never conflict with a user table. Although the `Information Schema` must be enabled by setting [`datafusion.catalog.information_schema`](https://datafusion.apache.org/user-guide/configs.html) to `true`. Also instead of having a separate `MemoryPool` in `StorageEngine`, all memory allocations in ModelarDB should probably be tracked by the [`MemoryPool` ](https://docs.rs/datafusion/latest/datafusion/execution/memory_pool/trait.MemoryPool.html) provided by Apache  DataFusion so ModelarDB can dynamically balance memory allocation for the entire system instead of storage and query processing separately. Also, since it is a `trait`, different strategies could be used depending on the use case, e.g., edge nodes may want to prioritize ingestion while cloud nodes prioritize query processing.

PR #285, opened 2025-02-11
Remove current implementation of metrics

CGodiksen:
As described in https://github.com/ModelarData/ModelarDB-RS/issues/284, we want to redesign how metrics are handled due to the current implementation being error prone, among other issues. This PR removes the current implementation as well as the action that was used to collect metrics.


crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -43,19 +42,14 @@ pub(super) struct CompressedDataManager {
     /// The compressed segments before they are saved to persistent storage. The key is the name of
     /// the model table the compressed segments represents data points for so the Apache Parquet
     /// files can be partitioned by table.
-    compressed_data_buffers: DashMap<String, CompressedDataBuffer>,
+    pub(super) compressed_data_buffers: DashMap<String, CompressedDataBuffer>,
```
skejserjensen, line 45: Unclear why `compressed_data_buffer` has become `pub(super)` after the removal of the metrics. Maybe document it in an answer to this comment so we keep that knowledge.

skejserjensen (approved):

PR #295, opened 2025-02-28
Implement Gorilla for float compression

aabduvakhobov:
This PR implements the support for compressing float data type by Gorilla compression method.  It uses ELF compression method's implementation of Gorilla for floats: https://github.com/Spatio-Temporal-Lab/elf

Currently, ModelarDB implements Gorilla compression method for doubles, which uses a higher number of bits for storing leading and trailing zeros. This PR ensures that ModelarDB uses only the necessary amount of bits for compressing floats, thus improving its compression.

### DIFF
--- a/crates/modelardb_compression/src/models/gorilla.rs
+++ b/crates/modelardb_compression/src/models/gorilla.rs
@@ -16,8 +16,8 @@
 //! Implementation of the Gorilla model type which uses the lossless compression method for
 //! floating-point values proposed for the time series management system Gorilla in the [Gorilla
 //! paper]. To adapt Gorilla model type for float compression, float implementation of Gorilla
-//! from ELF repository was used [ELF repository]. The compression method has been extended with support for 
-//! lossy compression by replacing values with the previous value if possible within the error bound. 
+//! from ELF repository was used [ELF repository]. The compression method has been extended with support for
+//! lossy compression by replacing values with the previous value if possible within the error bound.
 //! As this compression method compresses the values of a time series segment using XOR and a variable
 //! length binary encoding, aggregates are computed by iterating over all values in the segment.
 //!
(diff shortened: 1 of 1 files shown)


crates/modelardb_compression/src/models/gorilla.rs
```diff
@@ -110,8 +112,12 @@ impl Gorilla {
         } else {
             // Store each new value as its leading zero bits (if necessary) and
             // meaningful bits (all bits from the first to the last one bit).
-            let leading_zero_bits = value_xor_last_value.leading_zeros() as u8;
+            let mut leading_zero_bits = value_xor_last_value.leading_zeros() as u8;
             let trailing_zero_bits = value_xor_last_value.trailing_zeros() as u8;
+            // Avoid overflow of leading bits.
```
skejserjensen, line 117: Maybe add a newline before the comment, extend the comment to say "leading zero bits", and extend the comment to describe why the "leading zero bits" may overflow if this is not done.

crates/modelardb_compression/src/models/gorilla.rs
```diff
@@ -128,15 +134,20 @@ impl Gorilla {
...
                 self.compressed_values.append_a_one_bit();
                 self.compressed_values
-                    .append_bits(leading_zero_bits as u64, 5);
+                    .append_bits(leading_zero_bits as u64, 4);
 
                 let meaningful_bits =
                     models::VALUE_SIZE_IN_BITS - leading_zero_bits - trailing_zero_bits;
-                self.compressed_values
-                    .append_bits(meaningful_bits as u64, 6);
+                // 0 is used to denote 32 meaningful bits.
```
skejserjensen, line 144: Maybe add a newline before the comment.

crates/modelardb_compression/src/models/gorilla.rs
```diff
@@ -128,15 +134,20 @@ impl Gorilla {
...
+                if meaningful_bits == 32 {
```
skejserjensen, line 145: Is a branch needed? Could it not be computed as `meaningful_bits % 32` to avoid the branch?

skejserjensen (changes requested):
Thank you very much for the PR. Do you have results showing how much these changes impact ingestion time and compression for real-life data sets? I compared this PR to the current `main` branch and I did not see an improvement to the compression, but ingestion time seemed to increase by around 1.5-3%.

aabduvakhobov:
I did some tests to check if ELF's adaptation of Gorilla for float compression (EGF) improves ModelarDB's compression. I did tests using two proprietary datasets and a public dataset. 

Compression results showed EGF compresses slightly worse than ModelarDB's current implementation for lower error bounds like 0% and 0.01% (uses values and bit rewriting).  I've seen a slight improvement in compression for higher error bounds like 1% and 5% when Gorilla is combined with value and bit rewriting. However, with these error bounds, the number of values compressed by Gorilla gets very small due to PMC and Swing efficiency.

As for ingestion speed, I also noticed 1-3% decline with EGF even after addressing the comments in the PR thread. To conclude, I don't think it is worth to replace ModelarDB's current Gorilla implementation with EGF. Thus, I am closing this PR.

Issue #277, opened 2025-01-29
Use #[tokio::test] in the integration tests


CGodiksen:
Currently we create a runtime in the `TestContext` and then use that runtime manually in the tests. We do this because we cannot have two separate `tokio` runtimes in the same context. We should attempt to use the `tokio::test` decorator instead to avoid having to manually use the runtime in each test to run async functions.

PR #296, opened 2025-03-01
Use tokio test macro in integration tests

CGodiksen:
This PR implements https://github.com/ModelarData/ModelarDB-RS/issues/277 by updating the integration tests to use the `tokio::test` macro. Instead of using an explicit runtime provided in `TestContext`, the tests are now `async` and use `await` directly. 

`TestContext` has also been simplified so it now uses `tokio::Command` and `Command::kill_on_drop()` instead of manually killing the server when dropping.


crates/modelardb_server/tests/integration_test.rs
```diff
@@ -76,45 +75,42 @@ enum TableType {
...
             client,
         }
     }
 
-    /// Restart the server and reconnect the client.
-    fn restart_server(&mut self) {
-        Self::kill_child(&mut self.server);
-        self.server = Self::create_server(&self.temp_dir, self.port);
-        self.client = Self::create_client(&self.runtime, &mut self.server, self.port);
+    /// Restart the server on a new port and reconnect the client.
```
skejserjensen, line 104: Maybe add an additional sentence stating why a new port is used.

crates/modelardb_server/tests/integration_test.rs
```diff
@@ -369,102 +332,92 @@ impl TestContext {
...
-                Ok(RecordBatch::new_empty(schema))
-            } else {
-                // unwrap() is used as ? makes the return type Result<RecordBatch, ArrowError>>.
-                Ok(compute::concat_batches(&query_result[0].schema(), &query_result).unwrap())
-            }
-        })
+        if query_result.is_empty() {
+            Ok(RecordBatch::new_empty(schema))
+        } else {
+            // unwrap() is used as ? makes the return type Result<RecordBatch, ArrowError>>.
```
skejserjensen, line 380: Maybe delete the `unwrap()` comment as it does not seem to be used elsewhere in the file.

skejserjensen (approved):

crates/modelardb_server/tests/integration_test.rs
CGodiksen: Fixed.

crates/modelardb_server/tests/integration_test.rs
CGodiksen: Fixed.

chrthomsen (approved):

PR #217, opened 2024-09-27
Expose S3 and Azure contructors

skejserjensen:
Multiple types support operating on data in the local file system and remote object stores. However, arguments had to be serialized to bytes with a homemade encoding and passed to `try_from_connection_info()` as a `&[u8]` to use these types with a remote object store. This PR add `try_from_s3_configuration()` and `try_from_azure_configuration()` to simplify using the types with a remote object store. While implementing these methods, unnecessary use of `async` was found and removed, and methods operating on local paths where changed from `&str` to `std::path::Path` to make it clear that a local path must be passed. To differentiate between `std::path::Path` and `object_store::path::Path` and keep the code consistent is `std::path::Path` always imported as `std::path::Path as StdPath` even if  `object_store::path::Path` is not in the same scope.

PR #329, opened 2025-04-05
Support relative paths in modelardb_client

skejserjensen:
This PR fixes a bug in `modelardb_client` that prevents a relative path to a file with queries from being used. The problem was due to the use of `object_store` for local file paths as it explicitly does not allow [relative file paths](https://docs.rs/object_store/latest/object_store/path/struct.Path.html). Thus, it has been replaced with types from Rust's standard library. Like in #217, Rust's `Path` type is imported as `std::path::Path as StdPath` everywhere to keep the code consistent. Likewise, `PathBuf` is imported as `std::path::PathBuf as StdPathBuf` to be consistent with `Path`. In addition to fixing the bug, the PR also simplifies the code of `modelardb_client` as `object_store` could be removed. An additional check was also added to the simple command-line parameter parser to improve error messages. While looking at how local paths are implemented in `delta_lake.rs`, it was initially unclear that a path without a prefix was supported as `None` was surrounded by two `Some` in the `match` Thus, `None`  has been moved to the top to make it more clear.

### DIFF
--- a/crates/modelardb_client/src/main.rs
+++ b/crates/modelardb_client/src/main.rs
@@ -21,21 +21,20 @@ mod helper;
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::env::{self, Args};
-use std::io::{self, BufRead, IsTerminal, Write};
+use std::fs::File;
+use std::io::{self, BufRead, BufReader, IsTerminal, Write};
+use std::path::{Path as StdPath, PathBuf as StdPathBuf};
 use std::process;
 use std::sync::Arc;
 use std::time::Instant;
 
 use arrow::array::ArrayRef;
-use arrow::datatypes::{Schema, ToByteSlice};
+use arrow::datatypes::Schema;
 use arrow::ipc::convert;
 use arrow::util::pretty;
 use arrow_flight::flight_service_client::FlightServiceClient;
 use arrow_flight::{Action, Criteria, FlightData, FlightDescriptor, Ticket, utils};
 use bytes::Bytes;
-use object_store::local::LocalFileSystem;
-use object_store::path::Path;
-use object_store::{ObjectMeta, ObjectStore};
 use rustyline::Editor;
 use rustyline::history::FileHistory;
 use tonic::transport::Channel;
@@ -62,9 +61,6 @@ const TRANSPORT_ERROR: &str = "transport error: no messages received.";
 /// cannot be read.
 #[tokio::main]
 async fn main() -> Result<()> {
-    let current_dir = env::current_dir()?;
-    let local_file_system = LocalFileSystem::new_with_prefix(current_dir)?;
-
     // Parse the command line arguments.
     let args = env::args();
     if args.len() > 3 {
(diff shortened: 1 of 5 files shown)


crates/modelardb_client/src/main.rs
```diff
@@ -108,10 +102,15 @@ async fn parse_command_line_arguments(
...
-        if let Ok(query_file) = local_file_system.head(&arg_path).await {
+        let arg_path = &StdPath::new(arg.as_str());
+        if arg_path.exists() {
             // Assumes all files contains queries.
-            maybe_query_file = Some(query_file);
+            maybe_query_file = Some(arg_path.to_path_buf());
+        } else if arg.starts_with(['.', '/']) {
+            // Prevent missing files from being used as host.
+            Err(ModelarDbClientError::InvalidArgument(format!(
+                "{arg} does not exist"
```
CGodiksen, line 112:
```suggestion
                "{arg} does not exist."
```

crates/modelardb_client/src/main.rs
```diff
@@ -108,10 +102,15 @@ async fn parse_command_line_arguments(
     let mut maybe_query_file = None;
 
     for arg in args {
```
CGodiksen, line 104: We could consider changing how command line arguments are parsed so it is more similar to how it works for the server and manager. It seems odd that the order of the arguments does not matter and that you can give multiple of the same argument, with only the last being relevant.

CGodiksen (approved):

crates/modelardb_client/src/main.rs
skejserjensen: Fixed.

crates/modelardb_client/src/main.rs
skejserjensen: > We could consider changing how command line arguments are parsed so it is more similar to how it works for the server and manager.

It would certainly be simpler to require each argument to have a specific position. However, since it would require either the `host:port` or `query_file` to always be passed if the other is, I decided against it to make it simpler for the user. An alternative method could be to use flags such as `--host/-H host:port` and `--query-file/-f query_file`. If we use flags, it would be preferable to do so consistently for all binaries when arguments are optional instead of only doing it for `modelardb_client`.

> It seems odd that the order of the arguments does not matter...

Is it not a benefit for the user that the order does not matter if we can reliably distinguish between them? Of course, it does make the code more complex.

> ... and that you can give multiple of the same argument, with only the last being relevant.

True. We could limit the number of arguments to max two and return an error if the user passes the same argument multiple times to make it more clear that they should not do that.

What do you think @CGodiksen and @chrthomsen?

chrthomsen: Fine if solved as discussed today with three possible patterns: "A B", "A", "B"

chrthomsen (approved):

crates/modelardb_client/src/main.rs
```diff
@@ -62,72 +61,58 @@ const TRANSPORT_ERROR: &str = "transport error: no messages received.";
...
-
-    Ok((maybe_host, maybe_port, maybe_query_file))
+/// Parse the host and port in `maybe_host_port` and return a pair with the host of the server to
+/// connect to and the port to connect to. If the host is not included `DEFAULT_HOST` is used and if
+/// port is not included `DEFAULT_PORT` is used. Returns [`ModelarDbClientError`] if the port is not
+/// valid.
+fn parse_host_port(maybe_host_port: &str) -> Result<(&str, u16)> {
+    let mut parts = maybe_host_port.splitn(2, ':');
+    let host = parts.next().unwrap_or(DEFAULT_HOST);
+    let port = match parts.next() {
```
CGodiksen, line 108: Maybe use `Option::map_or()` instead of `match` here.

CGodiksen (approved):

crates/modelardb_client/src/main.rs
skejserjensen: Fixed.

Issue #260, opened 2024-11-25
Create a modelardb_test crate for test functionality 


CGodiksen:
To avoid duplicated code in the tests and to make organization of test functionality easier to handle, we should create a separate create for test code that can be used as a development dependency in the other crates. 

A discussion on this was started here: https://github.com/ModelarData/ModelarDB-RS/pull/259#discussion_r1855221632

PR #337, opened 2025-07-06
Create a new modelardb_test crate for test functionality

CGodiksen:
This PR implements https://github.com/ModelarData/ModelarDB-RS/issues/260 by creating a new `modelardb_test` crate for shared test functionality. `modelardb_common` has also been removed since most of the code that was left in `modelardb_common` after recent changes was test util.

Note that some functionality was duplicated when moved out of `modelardb_common` and into `modelardb_manager` and `modelardb_server`. This is not considered a problem since `modelardb_manager` will be removed.


crates/modelardb_test/Cargo.toml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 The ModelarDB Contributors
+# Copyright 2025 The ModelarDB Contributors
```
skejserjensen, line 1: Are the dates in the PR purposely updated? I do not recall us doing this in the past @CGodiksen and @chrthomsen.

skejserjensen (approved):

crates/modelardb_test/Cargo.toml
CGodiksen: Fixed

Issue #338, opened 2025-07-06
Change the CreateTables action to CreateTable


CGodiksen:
The `CreateTables` action should be changed to `CreateTable` to solve issues with the endpoint when trying to create multiple tables at the same time. For example, the current implementation can result in only some of the tables being created before failing and the request containing no tables. Furthermore, `CreateTable` is more consistent with the other actions such as  `RegisterNode` that also only manipulate a single element at a time.

PR #346, opened 2025-09-04
Change CreateTables to CreateTable

CGodiksen:
This PR implements https://github.com/ModelarData/ModelarDB-RS/issues/338 to make the API more consistent and to avoid issues with partial table creation. 

Since the old `TableMetadata` protobuf message was used for the response in `InitializeDatabase` as well, this PR also changes the start up flow for cluster nodes to no longer use `InitializeDatabase`. Now edge nodes use the remote data folder to initialize the local data folder with potentially missing tables. This change has already been discussed as part of the discussion on how to remove the manager. Therefore, it was determined that it was simpler to just do it now, instead of having to implement a temporary workaround.

### DIFF
--- a/crates/modelardb_server/src/manager.rs
+++ b/crates/modelardb_server/src/manager.rs
@@ -130,7 +130,7 @@ impl Manager {
                     .await?;
 
                 let schema = TableProvider::schema(&delta_table);
-                context.create_normal_table(&table_name, &schema).await?;
+                context.create_normal_table(table_name, &schema).await?;
             } else {
                 let time_series_table_metadata = remote_metadata_manager
                     .time_series_table_metadata_for_time_series_table(table_name)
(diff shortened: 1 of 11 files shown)


skejserjensen (changes requested):

skejserjensen (approved):

crates/modelardb_server/src/manager.rs
```diff
@@ -78,30 +77,70 @@ impl Manager {
...
-            Manager::new(flight_client, manager_metadata.key),
+            Manager::new(manager_metadata.key),
             manager_metadata.storage_configuration.unwrap(),
         ))
     }
 
     /// Initialize the local database schema with the normal tables and time series tables from the
-    /// managers database schema. If the tables to create could not be retrieved from the manager,
-    /// or the tables could not be created, return [`ModelarDbServerError`].
+    /// managers database schema using the remote data folder. If the tables to create could not be
```
chrthomsen, line 86: manager's

chrthomsen (approved):

PR #350, opened 2025-09-09
Add lossless error bound enum variant 

### LINKED ISSUE #345: Add lossless ErrorBound enum instance
The [`ErrorBound`](https://github.com/ModelarData/ModelarDB-RS/blob/66a4abf7f10f7a790dd195f0d962fe44183100cf/crates/modelardb_types/src/types.rs#L267) enum currently supports an `Absolute` and `Relative` enum. This means lossless compression is encoded as one of these enum instances with a value of `0.0`. In [PR](https://github.com/ModelarData/ModelarDB-RS/pull/341#discussion_r2265216952) #341 it was proposed to add a `Lossless` enum instance so it would be clear when lossless compression is done.

CGodiksen:
Closes #345 by adding a `Lossless` variant to the `ErrorBound` enum. The absolute and relative error bounds have also been changed both in the Rust API and in the Python bindings to no longer allow a value of 0.

When fixing some unit tests, a bug was found that allowed `TimeSeriesTableMetadata` to be created with unsupported types for specifically tag columns. This bug has been fixed and the test that checked for the issue has been fixed.

### DIFF
--- a/crates/modelardb_types/src/types.rs
+++ b/crates/modelardb_types/src/types.rs
@@ -268,13 +268,15 @@ fn compute_indices_of_columns_with_data_type(schema: &Schema, data_type: DataTyp
         .collect()
 }
 
-/// Absolute or relative per-value error bound.
+/// Absolute, relative, or lossless per-value error bound.
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub enum ErrorBound {
     /// An error bound that guarantees each value cannot deviate more than the [`Value`].
     Absolute(Value),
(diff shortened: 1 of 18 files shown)


crates/modelardb_types/src/types.rs
```diff
@@ -145,6 +146,21 @@ impl TimeSeriesTableMetadata {
             ));
         }
 
+        // If the schema contains an unsupported data type that does not correspond to a timestamp,
```
skejserjensen, line 149: Maybe extend the comment to explain how this can happen, I would assume the parser would prevent this?

PR #353, opened 2025-10-26
Replace unwrap() with expect("why")

### LINKED ISSUE #249: Replace unwrap() with expect("why") or Result?
Currently a few `unwrap()`s are used in non-test code when an `Option` is returned that cannot be `None` and when a `Result` is returned that cannot be `Err`. By convention, the use of `unwrap()` is accommodated by a comment that documents why it is safe to use. However, to simplify debugging in the case an assumption that would make an `unwrap()` safe does not hold, maybe `expect()` should be used instead with information about which assumption did not hold as the argument. Although, if it is safe for the process to continue running despite the assumption being wrong, an `Result` or an `Option` should probably be returned to allow the system to recover from the error instead of simply terminating. Also if possible due to the use of `unwrap()` in test code, maybe enable the [optional Clippy lints](https://rust-lang.github.io/rust-clippy/master) that detect of `unwrap()` and `panic()`.

CGodiksen:
Closes #249 by replacing most uses of `unwrap()` with `expect("why should it not fail")`. Note that we still use `unwrap()` in tests. The message style in the expect messages is based on the [Rust style guide](https://doc.rust-lang.org/std/error/index.html#common-message-styles).

### DIFF
--- a/crates/modelardb_bulkloader/src/main.rs
+++ b/crates/modelardb_bulkloader/src/main.rs
@@ -53,21 +53,13 @@ async fn main() -> Result<()> {
     }
 
     // Drop the path of the binary.
-    args.next()
-        .expect("args should contain at least five arguments.");
+    let expect_five_args = "args should contain at least five arguments.";
+    args.next().expect(expect_five_args);
 
-    let operation = args
-        .next()
-        .expect("args should contain at least five arguments.");
-    let input_path = args
-        .next()
-        .expect("args should contain at least five arguments.");
-    let output_path = args
-        .next()
-        .expect("args should contain at least five arguments.");
-    let table_name = args
-        .next()
-        .expect("args should contain at least five arguments.");
+    let operation = args.next().expect(expect_five_args);
+    let input_path = args.next().expect(expect_five_args);
+    let output_path = args.next().expect(expect_five_args);
+    let table_name = args.next().expect(expect_five_args);
 
     // Collect arguments for flags.
     let mut pre_sql: Vec<String> = vec![];
--- a/crates/modelardb_server/src/storage/types.rs
+++ b/crates/modelardb_server/src/storage/types.rs
@@ -24,6 +24,9 @@ use crate::error::Result;
 use crate::storage::compressed_data_buffer::CompressedSegmentBatch;
 use crate::storage::uncompressed_data_buffer::{IngestedDataBuffer, UncompressedDataBuffer};
 
+/// Message given when failing to acquire a mutex or wait on a mutex.
+const EXPECT_MUTEX_NOT_POISONED: &str = "Mutex should not be poisoned.";
+
 /// Resizeable pool of memory for tracking and limiting the amount of memory used by the
 /// [`StorageEngine`](crate::storage::StorageEngine). Signed integers are used to simplify updating
 /// the amount of available memory at runtime. By using signed integers for the amount of available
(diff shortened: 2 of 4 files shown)


crates/modelardb_bulkloader/src/main.rs
```diff
@@ -52,14 +52,22 @@ async fn main() -> Result<()> {
...
-    args.next().unwrap();
-
-    // unwrap() on args.next() is safe as args contains at least five elements.
-    let operation = args.next().unwrap();
-    let input_path = args.next().unwrap();
-    let output_path = args.next().unwrap();
-    let table_name = args.next().unwrap();
+    // Drop the path of the binary.
+    args.next()
+        .expect("args should contain at least five arguments.");
```
skejserjensen: Instead of duplicating the string, maybe assign it to a `str` variable and pass it to each.

crates/modelardb_bulkloader/src/main.rs
```diff
@@ -52,14 +52,22 @@ async fn main() -> Result<()> {
         print_usage_and_exit_with_error();
     }
 
-    // unwrap() is safe as the path of the binary is dropped.
```
skejserjensen: @CGodiksen and @chrthomsen open question, should we keep all the `unwrap()` comments and just replace `unwrap()` with `expect()`. Since the `expect()` messages seem to inform the user about what went wrong, while the comments seem to inform the developer what invariance should prevent the `expect()` from ever being triggered both seem useful.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -60,63 +60,80 @@ impl MemoryPool {
...
         compressed_memory_in_bytes: usize,
     ) -> Self {
-        // unwrap() is safe as i64::MAX is 8192 PiB and the value is from ConfigurationManager.
         Self {
             wait_for_ingested_memory: Condvar::new(),
             remaining_ingested_memory_in_bytes: Mutex::new(
-                ingested_memory_in_bytes.try_into().unwrap(),
+                ingested_memory_in_bytes
+                    .try_into()
+                    .expect("ingested_memory_in_bytes should be less than 8192 PiB."),
```
skejserjensen: Maybe use a variable instead of repeating the same string.

crates/modelardb_server/src/storage/types.rs
```diff
@@ -60,63 +60,80 @@ impl MemoryPool {
...
     }
 
     /// Change the amount of memory available for ingested data by `size_in_bytes`.
     pub(super) fn adjust_ingested_memory(&self, size_in_bytes: isize) {
-        // unwrap() is safe as lock() only returns an error if the mutex is poisoned.
-        *self.remaining_ingested_memory_in_bytes.lock().unwrap() += size_in_bytes;
+        *self
+            .remaining_ingested_memory_in_bytes
+            .lock()
+            .expect("Mutex should not be poisoned.") += size_in_bytes;
```
skejserjensen: Maybe use a variable or global constant instead of repeating the same string.

crates/modelardb_storage/src/query/sorted_join_exec.rs
```diff
@@ -253,8 +253,10 @@ impl SortedJoinStream {
     /// rows and if not drop rows from the [`RecordBatches`](RecordBatch) that contain extra. This
     /// can occur as compressed segments are not transferred atomically to the remote data folder.
     fn set_batch_num_rows_to_smallest(&mut self) {
-        // unwrap() is safe as a record batch is read from each input before this method is called.
-        let first_batch_num_rows = self.batches[0].as_ref().unwrap().num_rows();
+        let first_batch_num_rows = self.batches[0]
+            .as_ref()
+            .expect("A record batch should be read from each input before this method is called.")
```
skejserjensen: Maybe use a variable or a global constant instead of repeating the same string.

skejserjensen (changes requested):

crates/modelardb_server/src/storage/types.rs
CGodiksen: Fixed,

crates/modelardb_server/src/storage/types.rs
CGodiksen: Fixed.

Issue #368, opened 2025-12-01
Support persisting the node configuration to the Delta Lake


CGodiksen:
Currently the edge configuration is saved in memory and reset when the server is reset. This should be changed so it is saved to the Delta Lake and this table should be used when initializing the configuration manager.

PR #369, opened 2025-12-09
Save the configuration to a TOML configuration file

CGodiksen:
Closes https://github.com/ModelarData/ModelarDB-RS/issues/368 by updating the configuration manager so it now persists the configuration to a `modelardb.toml` configuration file in the root of the local data folder. This file is updated when changes are made to the configuration using the `UpdateConfiguration` action. If the file is changed manually, the changes are first applied when the server is restarted.

### DIFF
--- a/crates/modelardb_server/src/remote.rs
+++ b/crates/modelardb_server/src/remote.rs
@@ -641,7 +641,7 @@ impl FlightService for FlightServiceHandler {
     /// * `UpdateConfiguration`: Update a single setting in the configuration. The setting to update
     /// and the new value are provided in the [`UpdateConfiguration`](protocol::UpdateConfiguration)
     /// protobuf message in the action body. The setting is updated in the live server configuration
-    /// and the change is persisted in the `modelardbd.toml` configuration file.
+    /// and the change is persisted in the configuration file.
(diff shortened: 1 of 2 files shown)


chrthomsen (changes requested):

chrthomsen (approved):

crates/modelardb_server/src/remote.rs
```diff
@@ -640,7 +640,8 @@ impl FlightService for FlightServiceHandler {
     /// configuration is returned in a [`Configuration`](protocol::Configuration) protobuf message.
     /// * `UpdateConfiguration`: Update a single setting in the configuration. The setting to update
     /// and the new value are provided in the [`UpdateConfiguration`](protocol::UpdateConfiguration)
-    /// protobuf message in the action body.
+    /// protobuf message in the action body. The setting is updated in the live server configuration
+    /// and the change is persisted in the `modelardb.toml` configuration file.
```
skejserjensen: Update to `modelardbd.toml` if changed above. Alternatively, refer to `CONFIGURATION_FILE_NAME`, but it may be excessive to make it public just for referring to it in the documentation.

PR #376, opened 2026-03-13
Add first version of write-ahead log

CGodiksen:
This PR implements a write-ahead log (WAL) in modelardb_storage that ensures durability and crash recovery for ingested time series data. The WAL logs uncompressed data on a per-table basis before it enters the storage engine, and supports replaying unpersisted batches on startup to prevent data loss. The per-table WAL files are currently segmented based on batch count. Note that this PR includes the first version of the WAL, meaning that several future features have not been implemented yet.

The features that will be implemented in future PRs include controlling the threshold for segmentation and segmenting based on batch size, making it possible to disable the WAL, handling spilled buffers, explicitly handling data transfer and truncate, integration testing with fail-rs, and general optimizations.


crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -244,11 +253,22 @@ impl CompressedDataManager {
...
         self.local_data_folder
-            .write_compressed_segments_to_time_series_table(table_name, compressed_segments)
+            .write_compressed_segments_to_time_series_table(
+                table_name,
+                compressed_segments,
+                batch_ids.clone(),
+            )
             .await?;
 
+        // Inform the write-ahead log that data has been written to disk. We use a read lock since
```
skejserjensen: Unclear why an external log is even needed if internal locking is done.

crates/modelardb_server/src/storage/mod.rs
```diff
@@ -259,7 +260,30 @@ impl StorageEngine {
...
+            batch_id,
+        )
+    }
+
+    /// Pass `data_points` to [`UncompressedDataManager`] with a batch id given to the data by the
+    /// WAL. Return [`Ok`] if all of the data points were successfully inserted, otherwise return
+    /// [`ModelarDbServerError`].
+    pub(super) fn insert_data_points_with_batch_id(
+        &mut self,
+        time_series_table_metadata: Arc<TimeSeriesTableMetadata>,
```
skejserjensen: Maybe also use a type here to keep the arguments together to ensure they are for the same data and that the `RecordBatch`es have the correct schema.

crates/modelardb_server/src/context.rs
```diff
@@ -249,6 +268,40 @@ impl Context {
...
+        let local_data_folder = &self.data_folders.local_data_folder;
+
+        let write_ahead_log = self.write_ahead_log.write().await;
+        let mut storage_engine = self.storage_engine.write().await;
+
+        for metadata in local_data_folder.time_series_table_metadata().await? {
+            let unpersisted_batches =
+                write_ahead_log.unpersisted_batches_in_table_log(&metadata.name)?;
+
+            if !unpersisted_batches.is_empty() {
```
skejserjensen: Maybe add a comment about why this emits a warning, should `unpersisted_batches` not be empty for all tables if the system was shut down correctly?

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+use crate::WRITE_AHEAD_LOG_FOLDER;
+use crate::data_folder::DataFolder;
+use crate::error::{ModelarDbStorageError, Result};
+
+/// Folder containing the WAL files for the operations log.
+const OPERATIONS_LOG_FOLDER: &str = "operations";
+
+/// Number of batches to write to a single WAL segment file before rotating to a new one.
+const SEGMENT_ROTATION_THRESHOLD: u64 = 100;
```
skejserjensen: Please add the reasoning behind the use of this value so we can remember it later.

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+        // Since the std:fs API is used, the location must be a local path.
+        if location.contains("://") {
+            return Err(ModelarDbStorageError::InvalidState(format!(
+                "Write-ahead log location '{location}' is not a local path."
+            )));
+        }
+
+        let log_folder_path = PathBuf::from(location).join(WRITE_AHEAD_LOG_FOLDER);
+
+        std::fs::create_dir_all(&log_folder_path)?;
```
skejserjensen: Why use the full `std::fs::function_name` path for all functions in `std::fs` instead of importing `std::fs` and then using `fs::function_name`?

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+            .lock()
+            .expect("Mutex should not be poisoned.");
+
+        let mut all_batches = Vec::new();
+        for segment in closed_segments.iter() {
+            let batches = read_batches_from_path(&segment.path)?;
+            all_batches.extend((segment.start_id..=segment.end_id).zip(batches));
+        }
+
+        // Add the active segment's batches to the end of the list.
```
skejserjensen: Maybe "Append the active..."

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+    #[test]
+    fn test_reopen_and_append_continues_batch_ids() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let folder_path = temp_dir.path().join(TIME_SERIES_TABLE_NAME);
+        let metadata = table::time_series_table_metadata();
+
+        let batch = table::uncompressed_time_series_table_record_batch(10);
+
+        // Fill and rotate one segment.
```
skejserjensen: This is a general comment. Maybe always include the comment ", then drop." like above when an explicit drop is done using `{ }` to make it clear it is on purpose.

skejserjensen (changes requested):

crates/modelardb_server/src/storage/uncompressed_data_manager.rs
```diff
@@ -558,21 +532,23 @@ impl UncompressedDataManager {
...
-        let (memory_use, maybe_data_points, time_series_table_metadata) =
+        let (memory_use, maybe_data_points, time_series_table_metadata, batch_ids) =
             match uncompressed_data_buffer {
                 UncompressedDataBuffer::InMemory(mut uncompressed_in_memory_data_buffer) => (
                     uncompressed_in_memory_data_buffer.memory_size(),
                     uncompressed_in_memory_data_buffer.record_batch().await,
                     uncompressed_in_memory_data_buffer
                         .time_series_table_metadata()
                         .clone(),
+                    uncompressed_in_memory_data_buffer.batch_ids().clone(),
```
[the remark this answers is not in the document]
CGodiksen: Same reason we clone `time_series_table_metadata()` above. The methods return `&` and if we do not clone then we get "`uncompressed_in_memory_data_buffer` does not live long enough".

crates/modelardb_server/src/context.rs
CGodiksen: I am not sure I understand your comment. We emit a warning if `unpersisted_batches` is NOT empty, meaning we have to replay data. I think it makes sense to use a `warn!` log if we have to replay data due to the system not being shut down correctly.

crates/modelardb_storage/src/lib.rs
```diff
@@ -62,6 +63,9 @@ const TABLE_FOLDER: &str = "tables";
 /// The folder storing metadata in the data folders.
 const METADATA_FOLDER: &str = "metadata";
 
+/// The folder storing the write-ahead log in the data folders.
+pub const WRITE_AHEAD_LOG_FOLDER: &str = "_modelardb_log";
```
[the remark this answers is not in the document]
CGodiksen: It was inspired by the naming of `_delta_log` but I have changed it to `wal`.

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+impl WriteAheadLog {
+    /// Create a new [`WriteAheadLog`] that stores the log in the root of `local_data_folder` in
+    /// the [`WRITE_AHEAD_LOG_FOLDER`] folder. If the folder does not exist, it is created. If the
+    /// log could not be created, return [`ModelarDbStorageError`].
+    pub async fn try_new(local_data_folder: &DataFolder) -> Result<Self> {
+        // Create the folder for the write-ahead log if it does not exist.
+        let location = local_data_folder.location();
+
+        // Since the std:fs API is used, the location must be a local path.
```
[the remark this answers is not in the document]

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+    writer: StreamWriter<File>,
+    /// The batch id to give to the next batch of data. Monotonically increasing across rotations.
+    next_batch_id: u64,
+}
+
+impl ActiveSegment {
+    /// Create a new [`ActiveSegment`] in `folder_path` with the given `start_id` and `schema`.
+    /// If the file could not be created, return [`ModelarDbStorageError`].
+    fn try_new(folder_path: PathBuf, schema: &Schema, start_id: u64) -> Result<Self> {
+        let path = folder_path.join(format!("{start_id}-.wal"));
```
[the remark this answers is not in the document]
CGodiksen: Good idea, it seems that the [documentation](https://arrow.apache.org/docs/format/Columnar.html#ipc-streaming-format) recommends the `.arrows` file extension.

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+}
+
+/// The currently active WAL segment being written to. All fields are mutated together
+/// during rotation and are protected by the mutex in [`WriteAheadLogFile`].
+struct ActiveSegment {
+    /// Path to the active segment file.
+    path: PathBuf,
+    /// Batch id of the first batch written to this segment.
+    start_id: u64,
+    /// Writer to write data in IPC streaming format to this segment file.
```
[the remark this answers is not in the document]

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+    }
+
+    /// Read all data from all segment files (closed and active) in order and return them as pairs
+    /// of (batch_id, batch). If any file could not be read, return [`ModelarDbStorageError`].
+    fn read_all(&self) -> Result<Vec<(u64, RecordBatch)>> {
+        // Acquire the mutex to ensure data is not being written while reading.
+        let active = self
+            .active_segment
+            .lock()
+            .expect("Mutex should not be poisoned.");
```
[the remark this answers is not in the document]
CGodiksen: This expect message was chosen due to the convention started in `MemoryPool`. If this should be changed I think we should create an issue and do it everywhere after.

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -0,0 +1,1388 @@
...
+        Ok(self
+            .read_all()?
+            .into_iter()
+            .filter(|(batch_id, _)| !persisted.contains(batch_id))
+            .collect())
+    }
+
+    /// Read all data from all segment files (closed and active) in order and return them as pairs
+    /// of (batch_id, batch). If any file could not be read, return [`ModelarDbStorageError`].
+    fn read_all(&self) -> Result<Vec<(u64, RecordBatch)>> {
```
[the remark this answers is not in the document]

skejserjensen (approved):

chrthomsen (approved):

PR #378, opened 2026-03-17
Move modelardb_embedded definitions to C header

### LINKED ISSUE #373: Move modelardb_embedded definitions to C header
Currently, `modelardb_embedded`'s definitions are embedded into the `Operations` Python class. If they were in a separate C header, e.g., like in [TerseTS](https://github.com/cmcuza/TerseTS/blob/main/bindings/c/tersets.h), other languages that can use C headers would be able to use them and thus `modelardb_embedded`. For example, these include C, C++, Objective C, and Zig. If this change is made, the `Operations` class should be updated to read the C header and the Python package build scripts should be updated to include the C header in the package.

CGodiksen:
This PR closes #373 by moving modelardb_embedded definitions to a C header. The C header is read in `operations.py` similarly to how the library is read and the Python package build scripts have been updated to include the header file.

The PR also includes some minor fixes to consistency issues between `capi.rs` and the header file and a fix to a small bug found in the table Python tests.

One thing to note is that we include the `ARROW_C_DATA_INTERFACE` from https://arrow.apache.org/docs/format/CDataInterface.html directly in the header file. Based on the documentation it seems that the intention is to copy these definitions directly into your own projects code so this is what we do.

### DIFF
--- a/crates/modelardb_embedded/bindings/python/modelardb/operations.py
+++ b/crates/modelardb_embedded/bindings/python/modelardb/operations.py
@@ -23,8 +23,8 @@ from typing import Self
 
 import pyarrow
 from pyarrow import MapArray, RecordBatch, Schema, StringArray
-from pyarrow.cffi import ffi
 
+from .cffi import ffi
 from .error_bound import AbsoluteErrorBound
 from .table import NormalTable, TimeSeriesTable
 from .ffi_array import FFIArray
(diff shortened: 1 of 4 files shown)


crates/modelardb_embedded/bindings/python/modelardb/operations.py
```diff
@@ -111,116 +111,41 @@ def __find_library(build: str) -> str:
...
+            if not header_path.exists():
+                # Attempt to load the header from the development repository.
+                header_path = pathlib.Path(__file__).parent.parent.parent.resolve() / "c" / "modelardb_embedded.h"
+
+            if not header_path.exists():
+                raise FileNotFoundError("The C header modelardb_embedded.h was not found.")
+
+            content = header_path.read_text(encoding="UTF-8")
+
+            # Remove the Arrow C Data Interface block as pyarrow.cffi already defines it.
```
skejserjensen: Could the code removing preprocessor directives be simplified by iterating over it and skipping all lines from a `#ifdef ` to a `#endif# and any individual lines starting with `#include`? At the moment, it is a bit hard to follow what is removed and when. For example, something like using a `for` loop with an `if` statement with four branches: start skipping, end skipping, skip line, and append lines to the cleaned header.

skejserjensen (approved):

crates/modelardb_embedded/bindings/python/modelardb/operations.py
skejserjensen: Alternatively, maybe [`pycparser`](https://github.com/eliben/pycparser) can be used to remove the preprocessor directives that `cffi` does not handle. `pycparser` is already a dependency of [`cffi`](https://github.com/python-cffi/cffi/blob/main/pyproject.toml), so it would not add another dependency.

PR #382, opened 2026-03-20
Clean DataFolder and extend tests

### LINKED ISSUE #357: Clean DataFolder and extend its tests
Multiple types for managing tables through `object_store` were merged into `DataFolder` in #355. The tests from the other types were also moved, but now that all functionality is in one place, it should be possible to remove duplicate code by refactoring shared functionality into helper methods. For example, using the same code for managing `normal tables` and `metadata tables`. Also, as stated in #302, it would simplify the code if most methods worked on any table type instead of requiring special public methods for different table types. In addition, it may be necessary to update `DataFolder`'s tests to also cover the functionality that was added from the other types, as their existing tests were simply moved.

CGodiksen:
This PR closes #357 by cleaning up the API of `DataFolder` and extending the tests to cover all functionality. 

The main changes in the PR are:
- Combining functionality for handling metadata tables and normal tables
- Adding a new write method that can be used for both normal tables and time series tables
- Cleaning up the public API of `DataFolder`
- Reordering methods and tests to better match the rest of the repository
- Adding tests for all public methods in `DataFolder`
- Adding tests for all public methods in `DeltaTableWriter`

### DIFF
--- a/crates/modelardb_storage/src/data_folder/delta_table_writer.rs
+++ b/crates/modelardb_storage/src/data_folder/delta_table_writer.rs
@@ -115,7 +115,7 @@ impl DeltaTableWriter {
         };
 
         // A UUID version 4 is used as the operation id to match the existing Operation trait in the
-        // deltalake crate as it is pub(trait) and thus cannot be used directly in DeltaTableWriter.
+        // deltalake crate as it is pub(crate) and thus cannot be used directly in DeltaTableWriter.
         let operation_id = Uuid::new_v4();
 
         // Writer that will write the record batches.


crates/modelardb_storage/src/data_folder/mod.rs
```diff
@@ -862,65 +659,178 @@ impl DataFolder {
...
-            .await
+    /// Return a [`DeltaTableWriter`] for writing to the table with `table_name` in the Delta Lake,
+    /// or a [`ModelarDbStorageError`] if a connection to the Delta Lake cannot be established or
+    /// the table does not exist.
+    pub async fn table_writer(&self, table_name: &str) -> Result<DeltaTableWriter> {
+        let delta_table = self.delta_table(table_name).await?;
+        let partition_columns = delta_table.snapshot()?.metadata().partition_columns();
+
+        // If the table is a time series table, the partition column is the field column.
+        // self.is_time_series_table() is not used to avoid a redundant query to the metadata table.
```
skejserjensen: If this method for identifying if the table is a time seies table is more effiicent then the one currently implementted, maybe update `is_normal_table()` and `is_time_series_table()`.

PR #388, opened 2026-03-30
Fix bug where an empty folder is created when opening a Delta table that does not exist

### LINKED ISSUE #386: Fix side effect in delta_table_from_path() in DataFolder
`delta_table_from_path()` uses `deltalake::ensure_table_uri()` before `deltalake::open_table_with_storage_options()` which creates an empty folder if the folder for the table does not already exist. When `deltalake::open_table_with_storage_options()` then returns an error (because the table does not exist), the empty folder is left over.

CGodiksen:
This PR closes https://github.com/ModelarData/ModelarDB-RS/issues/386 by fixing the side effect in `DataFolder::delta_table_from_path()` that caused an empty folder to be created even when the Delta table did not exist due to the use of `deltalake::ensure_table_uri()`. 

This has been fixed and since `DataFolder::delta_table_from_path()` now works better with tables that does not exist, `is_normal_table` and `is_time_series_table` have been updated to use the Delta table to check for the table type. This avoids a query of the metadata table and is very efficient on repeated calls due to the Delta table cache in DataFolder.

### DIFF
--- a/crates/modelardb_storage/src/data_folder/mod.rs
+++ b/crates/modelardb_storage/src/data_folder/mod.rs
@@ -1639,6 +1639,14 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_missing_table_is_not_normal_table() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let data_folder = DataFolder::open_local(temp_dir.path()).await.unwrap();
+
+        assert!(!data_folder.is_normal_table("missing_table").await.unwrap());
+    }
+
+    #[tokio::test]
     async fn test_time_series_table_is_time_series_table() {
         let (_temp_dir, data_folder) = create_data_folder_and_create_time_series_table().await;
         assert!(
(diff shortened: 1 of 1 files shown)


crates/modelardb_storage/src/data_folder/mod.rs
```diff
@@ -762,7 +759,13 @@ impl DataFolder {
             Ok(delta_table.clone())
         } else {
             // If the table is not in the cache, open it and add it to the cache before returning.
-            let table_url = deltalake::ensure_table_uri(table_path)?;
+            let table_url =
+                deltalake::table::builder::parse_table_uri(table_path).map_err(|_error| {
```
skejserjensen, line 764: Unclear why this is manually mapped to an `InvalidArgument` instead of just using automatic wrapping by the question mark operator. If there is no implementation and/or conversion for the returned error type in `ModelarDbStorageError`, maybe implement it, if there is another reason why this is done manually, maybe add a comment to explain it.

skejserjensen (approved):

crates/modelardb_storage/src/data_folder/mod.rs
CGodiksen: Added a comment that specifies that we change it since the error message from deltalake is very unclear. Mapping it to our own error makes the error message of a few DataFolder methods much cleaner.

chrthomsen (approved):

PR #391, opened 2026-03-31
Segment WAL files based on a configurable maximum file size

### LINKED ISSUE #390: Make the WAL configurable by the user
Currently the `SEGMENT_BATCH_COUNT_THRESHOLD` is hard-coded to 100. This is not ideal for the scenario where a user wants more control over the WAL. Generally, basing the segment size on batch count is also not ideal since batches can be varying sizes.

This should all be changed so the user can configure the WAL and the segment size is based on something like row count instead.

CGodiksen:
This PR closes #390 by changing the WAL to segment files based on file size instead of batch count. This should make the segment files more consistent since batches can be hugely varying in size. Note that `RecordBatch::get_array_memory_size()` was chosen to estimate the size of the appended data for performance reasons. Other metrics such as the actual file size and row count were also considered but were not chosen due to low performance and low accuracy, respectively.

This PR also makes the maximum file size configurable and part of the configuration so it can be changed like any of our other configurations.


crates/modelardb_server/src/storage/compressed_data_manager.rs
```diff
@@ -572,7 +572,9 @@ mod tests {
             .unwrap();
 
         let write_ahead_log = Arc::new(RwLock::new(
-            WriteAheadLog::try_new(&local_data_folder).await.unwrap(),
+            WriteAheadLog::try_new(&local_data_folder, 64 * 1024)
```
skejserjensen: Maybe add a quick comment if `64 * 1024` is chosen for any specific reason.

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -648,6 +678,8 @@ mod tests {
     use modelardb_test::table::TIME_SERIES_TABLE_NAME;
     use tempfile::TempDir;
 
+    const SEGMENT_SIZE_THRESHOLD_IN_BYTES: u64 = 64 * 1024; // 64 KiB
```
skejserjensen: Missing . at the end of comment.

skejserjensen (approved):

crates/modelardb_storage/src/write_ahead_log.rs

chrthomsen (approved):

PR #394, opened 2026-04-23
Handle TRUNCATE operations in the WAL

### LINKED ISSUE #393: Handle TRUNCATE in the write-ahead log
It seems that the WAL already handles truncate since we flush the storage engine and therefore write all data to local disk and mark it as persisted before truncating the table. This should also truncate the WAL but there are a few issues that are not handled.

1. The active segment is not cleaned up in the WAL. Only closed segments are deleted when we mark batches as persisted.
2. If the Delta Lake transactions history is pruned, we lose what batch ids have been persisted and the WAL will replay truncated data.
3. If there are any missing batch ids, orphaned files will never be cleaned up.

To fix these issues we should explicitly handle truncate in the WAL.

CGodiksen:
This PR closes #393 by adding support for explicitly truncating a table log in the WAL. This is necessary to clean up active segments and any orphaned closed segments that were not deleted when the storage engine was flushed and the flushed data was marked as persisted. 

An explicity method to truncate also seems preferable due to the complexity of the interaction between the WAL and the Delta Lake. If we are explicit with all operations, it should be easier to keep them in sync.

### DIFF
--- a/crates/modelardb_server/src/context.rs
+++ b/crates/modelardb_server/src/context.rs
@@ -345,6 +345,7 @@ impl Context {
         // If the table is a time series table, truncate the table log file from the write-ahead
         // log to ensure data is not retained and replayed upon restart.
         if local_data_folder.is_time_series_table(table_name).await? {
+            // We use a read lock since the specific table log is locked internally before being truncated.
             let write_ahead_log = self.write_ahead_log.read().await;
             write_ahead_log.truncate_table_log(table_name)?;
         }
--- a/crates/modelardb_server/src/context.rs
+++ b/crates/modelardb_server/src/context.rs
@@ -329,8 +329,8 @@ impl Context {
     }
 
     /// Delete all data from the table with `table_name` if it exists. The table data is deleted
-    /// from the storage engine and Delta Lake. If the table does not exist or if it could not be
-    /// truncated, [`ModelarDbServerError`] is returned.
+    /// from the storage engine, write-ahead log, and Delta Lake. If the table does not exist or if
+    /// it could not be truncated, [`ModelarDbServerError`] is returned.
     pub async fn truncate_table(&self, table_name: &str) -> Result<()> {
         // Deleting the table from the storage engine does not require the table to exist, so the
         // table is checked first.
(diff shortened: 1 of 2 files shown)


crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -194,6 +194,14 @@ impl WriteAheadLog {
         table_log.unpersisted_batches()
     }
 
+    /// Truncate the table log for the given table name, removing all closed segment files and the
+    /// active segment file from disk, and starting a new active segment file. If the table log does
+    /// not exist or the data could not be truncated, return [`ModelarDbStorageError`].
+    pub fn truncate_table_log(&self, table_name: &str) -> Result<()> {
+        let table_log = self.table_log(table_name)?;
```
skejserjensen, line 201: Does `self.table_log(table_name)?.table_log.truncate()` cause compilation errors or is the use of `table_log` just easier to read?

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -563,6 +571,56 @@ impl SegmentedLog {
...
+
+        // Delete the active segment from disk.
+        debug!(
+            path = %active.path.display(),
+            "Deleting active WAL segment due to table truncate."
+        );
+
+        std::fs::remove_file(&active.path)?;
+
+        // Continue generating ids from the next unused batch id to avoid id collisions.
```
skejserjensen: Unclear how id collision can ever happen if all closed and active segments are deleted and the table is truncated? Why not just start from the lowest ID again? If this is not possible, then maybe expand the comment.

crates/modelardb_storage/src/write_ahead_log.rs
```diff
@@ -563,6 +571,56 @@ impl SegmentedLog {
...
+        let next_id = active.next_batch_id;
+        let new_active = ActiveSegment::try_new(self.folder_path.clone(), &self.schema, next_id)?;
+
+        // Commit the in-memory state transition only after all filesystem work succeeds.
```
skejserjensen: Maybe filesystem operations.

skejserjensen (approved):

crates/modelardb_storage/src/write_ahead_log.rs

crates/modelardb_storage/src/write_ahead_log.rs
CGodiksen: Expanded the comment to say that it can cause problems if the Delta Lake is restored by a rollback since the old batch ids can still be in the commit metadata.

crates/modelardb_storage/src/write_ahead_log.rs
CGodiksen: It is just for readability.

crates/modelardb_server/src/context.rs
chrthomsen, line 337: Not from this PR, but it is somewhat confusing when checking if a table exists that `Ok` means that it does not exist... Perhaps add a comment explaining this

chrthomsen (approved):

crates/modelardb_server/src/context.rs
CGodiksen: Renamed it to `check_table_does_not_exist` in the next PR.

PR #395, opened 2026-04-29
Make it possible to disable WAL

### LINKED ISSUE #392: Make it possible to disable the WAL
Currently the user is always forced to use the WAL since it is on be default with no options to turn it off. A environment variable should be added that makes it possible to disable the WAL if the user wants. It should still be enabled by default.

CGodiksen:
This PR closes https://github.com/ModelarData/ModelarDB-RS/issues/392 by making it possible to manually disable the WAL before the server is started either using an environment variable or by configuring it in the configuration file. This cannot be changed while the server is running.

We use a `WalMode` to manage the WAL now which is an enum that can be `Enabled(wal)` or `Disabled`. This matches the design of `ClusterMode`.


crates/modelardb_server/src/configuration.rs
```diff
@@ -348,6 +383,7 @@ impl ConfigurationManager {
             .await
     }
 
+    #[allow(dead_code)]
```
skejserjensen, line 386: Is this code only for tests? If so, maybe it can be conditionally compiled into only the test binaries using something like `'#[cfg(test)]'` to avoid it being marked as dead code. This would also make it clear what the code is for.

skejserjensen (approved):

crates/modelardb_server/src/configuration.rs
CGodiksen: Is it mostly for consistency with the other configurations, so they each have a getter and setter. However, I am fine with removing it if that is the consensus. What do you think @chrthomsen?

chrthomsen: OK, it sounds good to me to have getters/setters, but should the individual functions `test_...` below have been with `#[cfg(test)]` ?

chrthomsen (approved):

crates/modelardb_server/src/configuration.rs
CGodiksen: We have `#[cfg(test)]` on the module so the individual test functions do not need it.

Issue #299, opened 2025-03-05
Research issues with data order throughout the system


CGodiksen:
Due to how ingested data is sorted and how compressed data is sorted both when writing to disk and when reconstructing the data in the query engine, issues related to the order of the data might happen. 

One example of this is if out-of-order data points are added to the storage engine which then sorts and compresses them. This could cause overlapping compressed segments which could cause issues during querying. It is not certain that this causes issues but the problem should be researched to ensure this does not cause issues. A discussion on this was started here: https://github.com/ModelarData/ModelarDB-RS/pull/293#discussion_r1976508394.

Another potential issue is the disconnect between what Apache DataFusion expects from the ordering of the output from `SortedJoinExec` and what we currently provide. We currently only guarantee that the order is the same for each field column. It might be necessary to research what impact setting the sort guarantees and requirements has on the execution in Apache DataFusion and considering whether it would be better to remove the currently set guarantees. A discussion on this was started here: https://github.com/ModelarData/ModelarDB-RS/pull/293#discussion_r1976512756.

CGodiksen:
The data order might also be affected by issues with `optimize()`. If `optimize()` compacts one field more than another, one field can have more physical files than the others. Furthermore, since files are sorted by modification time, flushes and optimize can maybe cause ordering issues. A discussion on this was started here: https://github.com/ModelarData/ModelarDB-RS/pull/406#discussion_r3513820867

skejserjensen:
[...]

PR #406, opened 2026-07-01
Add support for manually optimizing a table

### LINKED ISSUE #239: Add support for using Delta Lake vacuum() and Delta Lake optimize() through Arrow Flight
Delta Lake has functionality for removing stale files and optimizing files, however since we provide a layer over Delta Lake, this functionality is not available to the user of ModelarDB. Actions should be added to Arrow Flight that makes it possible to use this functionality. 

When vacuuming or optimizing a table, the metadata Delta Lake should also be vacuumed or optimized since this is cheap and we do not want the user to have to manually vacuum or optimize the metadata Delta Lake.

- [ ] Add new endpoint to server Apache Arrow Flight server to vacuum a table.
- [ ] Add new endpoint to server Apache Arrow Flight server to optimize a table.
- [ ] Consider how to handle passing extra arguments such as retention period to endpoint.
- [ ] Add new endpoint to manager Apache Arrow Flight server to vacuum a table in the entire cluster.
- [ ] Add new endpoint to manager Apache Arrow Flight server to optimize a table in the entire cluster.
- [ ] Consider how to handle passing extra arguments to optimize endpoint.
- [ ] Consider if we should have endpoints to vacuum or optimize all tables.
- [ ] Add integration tests.
- [ ] Add unit tests.

CGodiksen:
This PR closes #239 by adding a new method to `DataFolder`, adding `OPTIMIZE [CLUSTER] [table_name[, table_name]+] [TARGET num_bytes]` to the SQL parser, adding a new method to `Cluster`, adding the structure in `Context` and the Apache Arrow Flight server to call optimize, and adding optimize to `modelardb_embedded` and the Python library. 

Unit testing have also been added in the relevant places and three new integration tests have been added to ensure the Apache Arrow Flight endpoint works. Note that the implementation matches `vacuum` 1-to-1.

### DIFF
--- a/crates/modelardb_embedded/bindings/python/tests/test_operations.py
+++ b/crates/modelardb_embedded/bindings/python/tests/test_operations.py
@@ -486,6 +486,41 @@ class TestOperations(unittest.TestCase):
             error_message = f"Invalid Argument Error: Table with name '{MISSING_TABLE_NAME}' does not exist."
             self.assertEqual(error_message, str(context.exception))
 
+    def test_data_folder_optimize(self):
+        with TemporaryDirectory() as temp_dir:
+            data_folder = Operations.open_local(temp_dir)
+            create_tables_in_data_folder(data_folder)
+
+            # Each write is a separate commit, so four writes produce multiple small files.
+            for _ in range(4):
+                data_folder.write(
+                    TIME_SERIES_TABLE_NAME,
+                    unsorted_time_series_table_data_without_generated_column(),
+                )
+
+            folder_path = os.path.join(temp_dir, "tables", TIME_SERIES_TABLE_NAME, "field_column=2")
+            file_count = len(os.listdir(folder_path))
+            self.assertEqual(file_count, 4)
+
+            data_folder.optimize(TIME_SERIES_TABLE_NAME)
+
+            # Vacuum to remove the compacted files.
+            data_folder.vacuum(TIME_SERIES_TABLE_NAME, retention_period_in_seconds=0)
+
+            # There should only be a single file left.
+            file_count = len(os.listdir(folder_path))
+            self.assertEqual(file_count, 1)
+
+    def test_data_folder_optimize_error(self):
+        with TemporaryDirectory() as temp_dir:
+            data_folder = Operations.open_local(temp_dir)
+
+            with self.assertRaises(RuntimeError) as context:
+                data_folder.optimize(MISSING_TABLE_NAME)
+
+            error_message = f"Invalid Argument Error: Table with name '{MISSING_TABLE_NAME}' does not exist."
+            self.assertEqual(error_message, str(context.exception))
+
 
 def create_tables_in_data_folder(data_folder: Operations):
     table_type = NormalTable(normal_table_schema())
--- a/crates/modelardb_embedded/bindings/python/tests/test_operations.py
+++ b/crates/modelardb_embedded/bindings/python/tests/test_operations.py
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Tests for the data folder operations exposed through the Python bindings.
+
+Test methods suffixed with ``_error`` only verify that errors from the
+underlying Rust implementation are propagated through the bindings and surface
+as Python exceptions. The errors themselves are already tested in the Rust
+implementation, so the specific error raised does not matter.
+"""
+
 import datetime
 import os
 import unittest
@@ -507,7 +515,7 @@ class TestOperations(unittest.TestCase):
             # Vacuum to remove the compacted files.
             data_folder.vacuum(TIME_SERIES_TABLE_NAME, retention_period_in_seconds=0)
 
-            # There should only be a single file left.
+            # The small files should be compacted into a single active file.
             file_count = len(os.listdir(folder_path))
             self.assertEqual(file_count, 1)
 
--- a/crates/modelardb_embedded/src/capi.rs
+++ b/crates/modelardb_embedded/src/capi.rs
@@ -1013,6 +1013,56 @@ unsafe fn vacuum(
     TOKIO_RUNTIME.block_on(modelardb.vacuum(table_name, maybe_retention_period_in_seconds))
 }
 
+/// Optimizes the table with the name in `table_name_ptr` in the [`DataFolder`] or [`Client`] in
+/// `maybe_operations_ptr` by compacting its many small files into fewer larger files of
+/// approximately `target_size_in_bytes_ptr` bytes. Assumes `maybe_operations_ptr` points to a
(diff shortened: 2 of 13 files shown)


crates/modelardb_embedded/src/capi.rs
```diff
@@ -963,11 +963,11 @@ unsafe fn drop(
 
 /// Vacuums the table with the name in `table_name_ptr` in the [`DataFolder`] or [`Client`] in
 /// `maybe_operations_ptr` by deleting stale files that are older than `retention_period_in_seconds_ptr`
-/// seconds. Assumes `maybe_operations_ptr` points to a [`DataFolder`] or [`Client`];
-/// `table_name_ptr` points to a valid C string; and `retention_period_in_seconds_ptr` points to a
-/// valid C string. A C string is used for the retention period to avoid issues with different
-/// platforms using an inconsistent amount of bits for integer types. The string is converted
-/// directly to an unsigned 64-bit integer in Rust.
+/// seconds. Assumes `maybe_operations_ptr` points to a [`DataFolder`] or [`Client`] and
+/// `table_name_ptr` points to a valid C string. `retention_period_in_seconds_ptr` points to a valid
```
chrthomsen: Isn't it still an assumption? Otherwise, perhaps "... must point to ... or be null to ..."?

crates/modelardb_storage/src/parser.rs
```diff
@@ -561,6 +578,78 @@ impl ModelarDbDialect {
...
+        } else {
+            vec![]
+        };
+
+        // If the next token is TARGET, attempt to parse the target file size in bytes.
+        let maybe_target_size_in_bytes = if let Token::Word(word) = parser.peek_nth_token(0).token
+            && word.keyword == Keyword::TARGET
+        {
+            parser.expect_keyword(Keyword::TARGET)?;
+            let target_size_in_bytes = self.parse_unsigned_literal_u64(parser)?;
```
chrthomsen, line 627: It is not strictly necessary, but it would be more user-friendly if K, M, and G suffixes could be used as in "TARGET 128M"

chrthomsen (approved):

crates/modelardb_embedded/src/capi.rs
CGodiksen: Fixed.

crates/modelardb_storage/src/parser.rs
CGodiksen: I could not find any built-in support for doing this in `sqlparser` but it should be possible to add manually. However I do not know if we should prioritize this. What do you think @skejserjensen?

crates/modelardb_embedded/bindings/python/tests/test_operations.py
```diff
@@ -486,6 +486,41 @@ def test_data_folder_vacuum_error(self):
...
+            data_folder.optimize(TIME_SERIES_TABLE_NAME)
+
+            # Vacuum to remove the compacted files.
+            data_folder.vacuum(TIME_SERIES_TABLE_NAME, retention_period_in_seconds=0)
+
+            # There should only be a single file left.
+            file_count = len(os.listdir(folder_path))
+            self.assertEqual(file_count, 1)
+
+    def test_data_folder_optimize_error(self):
```
skejserjensen, line 522: Maybe add something like `_missing_table` to the method name so it is clear what the error being tested is.

crates/modelardb_embedded/src/operations/data_folder.rs
```diff
@@ -2276,6 +2296,75 @@ mod tests {
...
+                .await
+                .unwrap();
+        }
+
+        let delta_table = data_folder.delta_table(NORMAL_TABLE_NAME).await.unwrap();
+        assert_eq!(delta_table.get_file_uris().unwrap().count(), 4);
+
+        data_folder.optimize(NORMAL_TABLE_NAME, None).await.unwrap();
+
+        // The small files should be compacted into a single active file.
```
skejserjensen, line 2316: What is meant by "active"? It was not written for the Python test, so maybe just delete it everywhere.

crates/modelardb_embedded/src/operations/data_folder.rs
```diff
@@ -2276,6 +2296,75 @@ mod tests {
...
+        }
+
+        let delta_table = data_folder
+            .delta_table(TIME_SERIES_TABLE_NAME)
+            .await
+            .unwrap();
+        assert_eq!(delta_table.get_file_uris().unwrap().count(), 8);
+
+        data_folder
+            .optimize(TIME_SERIES_TABLE_NAME, None)
```
skejserjensen, line 2341: Does this guarantee that the order of the segments stays the same for all fields?

skejserjensen (changes requested):

crates/modelardb_storage/src/parser.rs
skejserjensen: It seems like a nice-to-have that probably should be implemented as a separate parse function that can be reused. For example, something like `parse_storage_size_and_maybe_unit()` which assumes there will be an integer with an optional unit and returns the storage size parsed in the specified unit or bytes (default).

CGodiksen: I agree, i created an issue (https://github.com/ModelarData/ModelarDB-RS/issues/407) to handle this later.

crates/modelardb_embedded/bindings/python/tests/test_operations.py
CGodiksen: The actual error does not matter, all the `_error` tests in the Python tests are purely to test error propagation since the actual errors are already tested in the Rust implementation. I have added a module doc comment that makes this clear.

crates/modelardb_embedded/src/operations/data_folder.rs
CGodiksen: I added it to the Python test instead to be consistent. I think using `active` makes sense since this is the standard terminology that Delta Lake use to describe files that are not unused and stale.

crates/modelardb_embedded/src/operations/data_folder.rs
CGodiksen: I have added a comment to https://github.com/ModelarData/ModelarDB-RS/issues/239 to research this further. As far as i can see there might be some issues with ordering due to the modification time of the files and the potential of having a different number of files for each field. I think it holds but the previous implementation without `optimize()` is also very brittle. 

I think we should keep this for now so we can implement the data optimizer and finish A4. After that, we can then prioritize a general audit of data order in the system.

skejserjensen: Makes sense to, I have added a comment to #299 to start a broader disccussion about this.

skejserjensen (approved):

PR #410, opened 2026-07-15
Allow an optional unit when specifying bytes in SQL

### LINKED ISSUE #407: Allow an optional unit when specifying bytes in SQL
Currently we only allow a positive integer when, for example, specifying the target size of files when optimizing (`OPTIMIZE table_name TARGET 1024`). To make it easier for users, it should be possible to provide an optional unit that converts the given number automatically (`OPTIMIZE table_name TARGET 1024 MB`). 

A discussion on this was started here: https://github.com/ModelarData/ModelarDB-RS/pull/406#discussion_r3514104712.

Sigurd00:
This PR closes #407 by adding support for an optional unit (B, KB, MB, GB,  TB) after num_bytes in  [CLUSTER] [table_name[, table_name]+] [TARGET num_bytes[unit]], e.g. TARGET 1024 MB is automatically converted to bytes.

Unit testing have also been added in parser.rs covering each unit, case and the relevant error cases.

This is my first PR on the project, so let me know if I've missed anything terms of conventions or process.

### DIFF
--- a/crates/modelardb_storage/src/parser.rs
+++ b/crates/modelardb_storage/src/parser.rs
@@ -672,8 +672,9 @@ impl ModelarDbDialect {
     /// [`Self::parse_unsigned_literal_u64`]. If the [`Token`] after the number is a [`Token::Word`]
     /// matching a supported byte unit (B, KB, MB, GB, or TB, case-insensitive), it is consumed and
     /// the number is multiplied by the number of bytes the unit represents, e.g., `1 KB` is parsed
-    /// as `1024`. A [`ParserError`] is returned if the number cannot be parsed as a [`u64`], or if
-    /// multiplying the number by the unit does not fit in a [`u64`].
+    /// as `1024`. A [`ParserError`] is returned if the number cannot be parsed as a [`u64`], if the
+    /// number is followed by a word that is not a supported byte unit, or if multiplying the number
+    /// by the unit does not fit in a [`u64`].
     fn parse_unsigned_literal_u64_with_optional_byte_unit(
         &self,
         parser: &mut Parser,
--- a/crates/modelardb_storage/src/parser.rs
+++ b/crates/modelardb_storage/src/parser.rs
@@ -2436,6 +2436,16 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


crates/modelardb_storage/src/parser.rs
```diff
@@ -2316,6 +2365,76 @@ mod tests {
...
+    #[test]
+    fn test_tokenize_and_parse_optimize_with_target_size_and_mixed_case_unit() {
+        let (table_names, maybe_target_size_in_bytes, cluster) =
+            parse_optimize_and_extract_table_names("OPTIMIZE TARGET 1 Mb");
+
+        assert!(table_names.is_empty());
+        assert_eq!(maybe_target_size_in_bytes, Some(1024 * 1024));
+        assert!(!cluster)
+    }
+
```
CGodiksen, line 2430: A test could be added for specifying the unit without a space between the number and the unit. This should be allowed.

CGodiksen (changes requested):
Thank you for the well-written PR. My apologies for the late review but we have been on summer vacation for the last 3 weeks. Regarding the PR, everything looks good and follows the conventions of the repository. The check currently fails due to `cargo clippy` but that will be fixed in https://github.com/ModelarData/ModelarDB-RS/pull/412 so don't worry about that in this PR. 

The syntax you added looks good, I just want one extra test to ensure that the space between the number and the unit is actually optional (which is what we want). Thank you again for your interest in our project!

crates/modelardb_storage/src/parser.rs
Sigurd00: I have added a test for this

CGodiksen (approved):
Thank you. The change suggested by Copilot also makes sense and should be applied before the PR is merged.

skejserjensen (approved):
LGTM, once again thank you for the PR.

skejserjensen:
@Sigurd00 and @CGodiksen #412 have now been merged, so if the branch is rebased onto `main`, `clippy` should be happy again :)

CGodiksen:
The PR has now been merged, thank you for your contribution @Sigurd00!

Sigurd00:
Thank you and thanks for the time to review :)

PR #412, opened 2026-08-01
Object store macros

### LINKED ISSUE #278: Simplify testing with different object store using macros
Currently, all tests that use `ObjectStore` use `LocalFileSystem` as it is always available. However, the tests should also be executed on the other types of `ObjectStore` to ensure they all work as expected by passing the tests. This could be implemented as a function that accepts an `ObjectStore` and performs the actual tests and which is then called by test functions that each create a different type of `ObjectStore`. However, this would require a significant amount of boilerplate code for each test and a new test function must added to all of the tests that use an `ObjectStore` every time support for a new type of `ObjectStore` is added. Inspired by [`#[tokio::test]`](https://docs.rs/tokio/latest/tokio/attr.main.html), a [Procedural Macro](https://doc.rust-lang.org/reference/procedural-macros.html) should be created to automatically generate a test for each type of `ObjectStore`, e.g., using an [Attribute Macro](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros). An open question is if the macro should check if the object stores are available when compiling so the tests can still be run on platforms without all of the object stores available.

skejserjensen:
This PR fixes #278 by adding two procedural macros, `data_folder_test` and object_store_test that generate `#[tokio::test]` functions that call the annotated function with all permutations with replacements of `DataFolder` configurations and `ObjectStore`'s, respectively. For example, the following code executes the annotated function with all permutations with replacements of `ObjectStore` can be used to call `test_object_store_list()` and the test completes successfully. Both macros also work with two and three parameters, and with more parameters, the macros generate more test functions as more are needed to test all permutations with replacements.

```rust
use modelardb_macros::object_store_test;

#[object_store_test]
async fn test_object_store_list(store_one: &dyn ObjectStore) {
    let mut files = store_one.list(None);
    while let Some(f) = files.next().await {
       assert!(......);
    }
}
```

The current implementation has a few known limitations that I deferred until we gain more experience with using it so we can improve it based on our experience instead of my guesses about how it might be used.
1) There are no automatic tests as it requires additional dependencies, and while most online recommend implementing Rust procedural macros using a set of extra crates that is not included with Rust, they made the code more complicated in my opinion. So since these macros are fairly simple as soon as one understands how Rust procedural macros work, I [decided to remove all the extra dependencies](https://github.com/ModelarData/ModelarDB-RS/commit/7d27d207fba5c4e1c8061050529361fa9597efec) and manually test the macros thoroughly before creating the PR.
2) I could not find a way to use `data_folder_test` to test `DataFolder` as it causes Rust to become confused about which version of the `DataFolder` struct to use. I tried to add a method to `DataFolder` so one can be created from an existing `ObjectStore` so we could implement all tests using only `object_store_test` and solve the problem that way; however, `DataFolder` performs initialization for `LocalFileSystem` that I could not figure out a way to implement in a method that accepts any `&dyn ObjectStore` and creates a correctly configured `DataFolder` from it. Another benefit of adding this method would also be to decouple the initialization of `ObjectStore` and `DataFolder` as `DataFolder` currently does both and thus has many `open_*()` methods.
3) `data_folder_test` does not work with Azurite, but `object_store_test` does, and as far as I can tell, the code is the same for both. However, since the `DataFolder` configuration for Azure explicitly states that it needs to be tested and thus probably needed to be changed, I left this for another PR to avoid making changes to large changes to `DataFolder` in this.


crates/modelardb_macros/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition = "2024"
+authors = ["Soeren Kejser Jensen <devel@kejserjensen.dk>"]
+
+[lib]
+proc-macro = true
+
+[dependencies]
+itertools.workspace = true
+modelardb_test = { path = "../modelardb_test" }
+tempfile.workspace = true
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_storage/src/data_folder/mod.rs
```diff
@@ -237,6 +239,7 @@ impl DataFolder {
             ("azure_storage_account_name".to_owned(), account_name),
             ("azure_storage_account_key".to_owned(), access_key),
             ("azure_container_name".to_owned(), container_name),
+            ("azure_storage_use_emulator".to_owned(), use_emulator),
         ]);
         let url = Url::parse(&location)
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! The procedural macros used throughout ModelarDB. The procedural macros does purposely not use
+//! crates designed to simply writing procedural macros like `syn`, `proc_macro2`, and `quote`, as
+//! they made the code more complex when evaluated. The macros have no automatic tests as using
+//! [`proc_macro`] outside procedural macros makes the compiler panic.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+        ParameterType::ObjectStore => {
+            expect_ident_with_contents(token_iterator, "dyn")?;
+            expect_ident_with_contents(token_iterator, "ObjectStore")?;
+        }
+    }
+    Ok(())
+}
+
+/// Return [`Ok`] if the next [`TokenTree`] from `token_iterator` is an [`Ident`], otherwise a
+/// [`ModelarDbMacrosError] is returned.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+    parameter_type: ParameterType,
+) -> (String, u16) {
+    let (function_name_ident, function_parameter_group) =
+        next_ident_and_group(input.clone()).expect("Assumes input is a function with parameters.");
+    let function_name = function_name_ident.to_string();
+    let data_folder_parameter_count =
+        expect_parameter_type_and_count(function_parameter_group, parameter_type)
+            .expect("Assumes all of the function's parameters are of type {parameter_type}.");
+    (function_name, data_folder_parameter_count)
+}
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+            "
+            #[tokio::test]
+            async fn {function_name}__{name}() {{
+                {function_name}({arguments}).await
+            }}
+        "
+        ));
+    }
+
+    println!("{}", code);
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    append_code_to_token_stream(input, code)
+}
+
+/// Extracts the function name from `input`, checks that all its parameters are of type
+///`ParameterType`, and returns the number of parameters. `The number of parameters is returned as a
+///`u16` as `rustc` returns an error if a function or method have more than 65,535 parameters at the
+///time of writing. An [`ModelarDbMacrosError`] is returned if `input` is not a function that only
+///accept parameters of `parameter_type`.
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_test/src/object_store.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Return an [`InMemory`] [`ObjectStore`] for testing.
+pub fn in_memory_object_store() -> Box<dyn ObjectStore> {
+    Box::new(InMemory::new())
+}
+
+/// Return a [`LocalFileSystem`] [`ObjectStore`] for testing.
+pub fn local_file_system_object_store() -> Box<dyn ObjectStore> {
+    let temp_dir = tempfile::tempdir().unwrap();
+    let local_file_system = LocalFileSystem::new_with_prefix(temp_dir.path()).unwrap();
+    Box::new(local_file_system)
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_test/src/data_folder.rs
```diff
@@ -0,0 +1,56 @@
...
+/// Return a [`DataFolder`] storing data in memory for testing.
+pub async fn in_memory_data_folder() -> DataFolder {
+    DataFolder::open_memory().await.unwrap()
+}
+
+/// Return a [`DataFolder`] storing data on local disk for testing.
+pub async fn local_file_system_data_folder() -> DataFolder {
+    let temp_dir = tempfile::tempdir().unwrap();
+    DataFolder::open_local(temp_dir.path()).await.unwrap()
+}
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+            .join(", ");
+
+        code.push_str(&format!(
+            "
+            #[tokio::test]
+            async fn {function_name}__{name}() {{
+                {function_name}({arguments}).await
+            }}
+        "
+        ));
```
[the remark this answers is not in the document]
skejserjensen: It tested the current code manually, and Rust could compile it without any errors or warnings.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+            // Clone is needed to avoid a partial move error with the next if let.
+            return Some((ident, group.clone()));
+        } else if let Ident(ident) = token {
+            adjacent_ident = Some(ident);
+        } else {
+            adjacent_ident = None;
+        }
+    }
+    None
+}
```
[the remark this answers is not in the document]
skejserjensen: This does not seem to be relevant functionality for our intended use case.

crates/modelardb_test/Cargo.toml
```diff
@@ -24,4 +24,7 @@ arrow.workspace = true
 async-trait.workspace = true
 datafusion.workspace = true
 modelardb_types = { path = "../modelardb_types" }
+modelardb_storage = { path = "../modelardb_storage" }
+object_store = { workspace = true, features = ["aws", "azure"] }
```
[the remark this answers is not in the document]
skejserjensen: Tested and executed without any issues.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+impl Display for ParameterType {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            ParameterType::DataFolder => write!(f, "&DataFolder"),
+            ParameterType::ObjectStore => write!(f, "&dyn ObjectStore"),
+        }
+    }
+}
+
+/// An function that will create an argument to be passed to a function as a borrow.
```
CGodiksen:
```suggestion
/// A function that will create an argument to be passed to a function as a borrow.
```

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+}
+
+/// Macro for generating test functions that use all permutations with replacements of `DataFolder`
+/// The macro must be placed on an `async` function without `#[test]` or `#[tokio::test]` that only
+/// has `&DataFolder` parameters. It will generate one `#[tokio::test]` function for each
+/// permutation with replacement of `DataFolder` configurations that call the annotated function.
+#[proc_macro_attribute]
+pub fn data_folder_test(
```
CGodiksen, line 90: Maybe add a small code example in the doc comment since we do that with our other macros.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+    // Append the generated functions to the existing token stream.
+    append_code_to_token_stream(input, code)
+}
+
+/// Macro for generating test functions that use all permutations with replacements of object
+/// stores. The macro must be placed on an `async` function without `#[test]` or `#[tokio::test]`
+/// that only has `&dyn ObjectStore` parameters. It will generate one `#[tokio::test]` function for
+/// each permutation with replacement of supported object stores that call the annotated function.
+#[proc_macro_attribute]
+pub fn object_store_test(
```
CGodiksen, line 177: Maybe add a small code example in the doc comment since we do that with our other macros.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+        function_name_and_checked_parameter_count(input.clone(), ParameterType::DataFolder);
+
+    // Generate the async tokio::test functions that will call the annotated function using all
+    // combinations of data folder configurations. A separate function is created for each
+    // permutation of data folder configurations instead of a single function with nested loops to
+    // make it simpler to see which combination of data folder configurations fail a test. The name
+    // for the generated function start with the name of the annotated function followed by the name
+    // of each data folder configuration used with each part of the name separated by two
+    // underscores. By using the annotated function name as a prefix, all the generated tests can be
+    // run with cargo test annotated function name. The names of the data folder configurations used
```
CGodiksen: Should this be "run with `cargo test annotated_function_name`." or something similar to make it clear that it is a command?

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+            true,
+        ),
+        BorrowedArgument::new(
+            "local_file_system_data_folder",
+            "modelardb_test::data_folder",
+            true,
+            true,
+        ),
+        BorrowedArgument::new(
+            "aws3_data_folder",
```
CGodiksen: Maybe `s3_data_folder` to be consistent with `DataFolder::open_s3`.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+        ),
+        BorrowedArgument::new(
+            "azure_data_folder",
+            "modelardb_test::data_folder",
+            false,
+            true,
+        ),
+    ];
+
+    let data_folder_permutations_with_replacements =
```
CGodiksen, line 139: Maybe add a comment to explain the logic here.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+            false,
+        ),
+        BorrowedArgument::new(
+            "local_file_system_object_store",
+            "modelardb_test::object_store",
+            true,
+            false,
+        ),
+        BorrowedArgument::new(
+            "aws3_object_store",
```
CGodiksen: If it is changed above, also change it here.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+        ),
+        BorrowedArgument::new(
+            "azure_object_store",
+            "modelardb_test::object_store",
+            false,
+            false,
+        ),
+    ];
+
+    let object_store_permutations_with_replacements =
```
CGodiksen, line 212: if a comment is added above, also add it here.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+        } else if let Ident(ident) = token {
+            adjacent_ident = Some(ident);
+        } else {
+            adjacent_ident = None;
+        }
+    }
+    None
+}
+
+/// Returns the number of `parameter_type` parameters in `function_parameter_group`. `The return
```
CGodiksen:
```suggestion
/// Returns the number of `parameter_type` parameters in `function_parameter_group`. The return
```

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+/// type is `u16` as `rustc` returns an error if a function or method have more than 65,535
+/// parameters at the time of writing. An [`ModelarDbMacrosError`] is returned if
```
CGodiksen:
```suggestion
/// parameters at the time of writing. A [`ModelarDbMacrosError`] is returned if
```

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+    Ok(parameter_count)
+}
+
+/// Returns [`Ok`] if the next four tokens `token_iterator` returns is as expected according to
+/// `parameter_type`, otherwise a [`ModelarDbMacrosError`] is returned.
+fn expect_parameter_type(
+    token_iterator: &mut impl Iterator<Item = TokenTree>,
+    parameter_type: ParameterType,
+) -> Result<()> {
+    // The contents of the first Ident token cannot checked as it is the parameter name.
```
CGodiksen:
```suggestion
    // The contents of the first Ident token cannot be checked as it is the parameter name.
```

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+) -> Result<()> {
+    let error_message = match token_iterator.next() {
+        Some(Ident(token)) if token.to_string() == contents => return Ok(()),
+        Some(token) => format!("Expected Ident with {}, found {}.", contents, token),
+        None => format!("Expected Ident with {}, found an empty iterator.", contents),
+    };
+    Err(ModelarDbMacrosError::Parse(error_message))
+}
+
+/// Return [`Ok`] if the next [`TokenTree`] from `token_iterator` is an [`Punct`] that contains
```
CGodiksen:
```suggestion
/// Return [`Ok`] if the next [`TokenTree`] from `token_iterator` is a [`Punct`] that contains
```

crates/modelardb_macros/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+license = "Apache-2.0"
+edition = "2024"
+authors = ["Soeren Kejser Jensen <devel@kejserjensen.dk>"]
+
+[lib]
+proc-macro = true
+
+[dependencies]
+itertools.workspace = true
+modelardb_test = { path = "../modelardb_test" }
```
CGodiksen: This is not specific to this PR but did we decide to explicitly not use `.workspace = true` for local crate dependencies? It makes it easier to see what is local but we could also be consistent.

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,379 @@
+/* Copyright 2026 The ModelarDB Contributors
```
CGodiksen, line 1: Is it intentional that we still have normal macros in `macros.rs` in `modelardb_types` instead of having them in this crate? If they are moved, maybe have separate files for these macros and the moved macros to avoid having a single large file.

crates/modelardb_test/src/data_folder.rs
```diff
@@ -0,0 +1,57 @@
...
+pub async fn in_memory_data_folder() -> DataFolder {
+    DataFolder::open_memory().await.unwrap()
+}
+
+/// Return a [`DataFolder`] storing data on local disk for testing.
+pub async fn local_file_system_data_folder(data_folder_path: &StdPath) -> DataFolder {
+    DataFolder::open_local(data_folder_path).await.unwrap()
+}
+
+/// Return a [`DataFolder`] storing data in an AWS3 compatible object store for testing using Minio.
```
CGodiksen:
```suggestion
/// Return a [`DataFolder`] storing data in an Amazon S3-compatible object store for testing using MinIO.
```

crates/modelardb_test/src/data_folder.rs
```diff
@@ -0,0 +1,57 @@
...
+pub async fn aws3_data_folder() -> DataFolder {
```
CGodiksen: If this was changed in the `modelardb_macros` crate, also change it here.

crates/modelardb_test/src/data_folder.rs
```diff
@@ -0,0 +1,57 @@
...
+        "http://localhost:9000".to_owned(),
+        BUCKET_AND_CONTAINER_NAME.to_owned(),
+        "minioadmin".to_owned(),
+        "minioadmin".to_owned(),
+    )
+    .await
+    .unwrap()
+}
+
+/// Return a [`DataFolder`] storing data in Microsoft Azure for testing using Azurite.
```
CGodiksen:
```suggestion
/// Return a [`DataFolder`] storing data in Microsoft Azure Blob Storage for testing using Azurite.
```

crates/modelardb_test/src/lib.rs
```diff
@@ -15,9 +15,14 @@
 
 //! Implementation of functions and types used throughout ModelarDB for testing purposes.
 
+pub mod data_folder;
 pub mod data_generation;
+pub mod object_store;
 pub mod table;
 
+/// Bucket and container name used by Minio and Azurite.
```
CGodiksen:
```suggestion
/// Bucket and container name used by MinIO and Azurite.
```

crates/modelardb_test/src/object_store.rs
```diff
@@ -0,0 +1,88 @@
...
+    Box::new(InMemory::new())
+}
+
+/// Return a [`LocalFileSystem`] [`ObjectStore`] for testing.
+pub fn local_file_system_object_store(object_store_path: &StdPath) -> Box<dyn ObjectStore> {
+    let local_file_system = LocalFileSystem::new_with_prefix(object_store_path).unwrap();
+    Box::new(local_file_system)
+}
+
+/// Return a [`AmazonS3`](object_store::aws::AmazonS3) [`ObjectStore`] for testing.
```
CGodiksen:
```suggestion
/// Return an [`AmazonS3`](object_store::aws::AmazonS3) [`ObjectStore`] for testing.
```

CGodiksen (changes requested):

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed, although the code was not my idea; it was stolen directly from the [itertools documentation](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.permutations).

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: I have added a comment to the start of `object_store_test` that directs the reader to `data_folder_test` for detailed information about how both macros are implemented. This is so we do not have to maintain multiple large comments for both macros.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_macros/Cargo.toml
skejserjensen: Personally, I just never thought of it, but it seems like a good idea, so I have updated all our dependencies in our crates to always refer to the workspace.

crates/modelardb_macros/src/lib.rs
skejserjensen: Yes, it is intentional, as doing so currently results in this error `error: cannot export macro_rules! macros from a proc-macro crate type currently`. Unfortunately, procedural macros impose quite a few restrictions on their implementations, e.g., they must be implemented in the [root of a crate of type `proc-macro`](https://doc.rust-lang.org/reference/procedural-macros.html). This is why both macros are implemented in `lib.rs`. Of course, if we add more macros, we may consider moving the helper functions to other files, but the macros themselves must be in the crate root. The restrictions are probably due to [`proc-macro` crates](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-proc-macro-field) being [used by the compiler](https://doc.rust-lang.org/book/ch20-05-macros.html#procedural-macros-for-generating-code-from-attributes) and get access to the compiler information through the [`proc_macro`](https://doc.rust-lang.org/proc_macro/) crate, similar to OCaml's [`PPX`](https://ocaml.org/docs/metaprogramming). For more information about procedural macros, the best resources I have found are [The Little Book of Rust Macros](https://lukaswirth.dev/tlborm/) and Rust's own documentation. I found that most online blogs, tutorials, etc skips a lot of relevant information about how macros work and use a lot of 3rd-paty dependencies like [`quote`](https://docs.rs/quote/latest/quote/), [`proc_macro2`](https://docs.rs/proc-macro2/latest/proc_macro2/), and [`syn`](https://docs.rs/syn/latest/syn/) that made the code more complex when I tried to use them. The [`zyn` crate](https://github.com/aacebo/zyn) looks promising, but the crate's author has stated that `zyn` is just a prototype and will be [replaced by a new framework named `moxy`](https://github.com/aacebo/zyn/discussions/28). So in general, Rust macros are quite painful no matter the type, and it having multiple types of macros with different syntax and/or rules just makes the problem worse. So if anybody needs to implement software that requires significant compile-time code generation and they get to choose the implementation language, I would probably recommend [Zig](https://ziglang.org/documentation/0.16.0/) due to its [`comptime`](https://ziglang.org/documentation/0.16.0/#comptime) instead of Rust.

crates/modelardb_test/src/data_folder.rs
skejserjensen: Fixed.

crates/modelardb_test/src/data_folder.rs
skejserjensen: Fixed.

crates/modelardb_test/src/data_folder.rs
skejserjensen: Fixed.

crates/modelardb_test/src/lib.rs
skejserjensen: Fixed.

crates/modelardb_test/src/object_store.rs
skejserjensen: Fixed.

crates/modelardb_test/Cargo.toml
```diff
@@ -23,5 +23,8 @@ authors = ["Soeren Kejser Jensen <devel@kejserjensen.dk>"]
 arrow.workspace = true
 async-trait.workspace = true
 datafusion.workspace = true
-modelardb_types = { path = "../modelardb_types" }
+modelardb_types.workspace = true
+modelardb_storage.workspace = true
```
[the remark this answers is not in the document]
skejserjensen: Let's leave this for now until we know how we want to use and/or restructure the macros.

crates/modelardb_macros/src/lib.rs
CGodiksen: I think you "fixed" the wrong line here. In the commit you changed "an Ident" to "a Ident" but you did not change this line. It should also be changed back to "an Ident" in the doc comment for `expect_ident_without_contents`.

CGodiksen (approved):

crates/modelardb_macros/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+impl Display for ParameterType {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            ParameterType::DataFolder => write!(f, "&DataFolder"),
+            ParameterType::ObjectStore => write!(f, "&dyn ObjectStore"),
+        }
+    }
+}
+
+/// A function that will create an argument to be passed to a function as a borrow.
```
chrthomsen: But this is not a function? Or will the comment apply to the only function that is implemented further below?

chrthomsen (approved):

crates/modelardb_macros/src/lib.rs
skejserjensen: The doc comments describe what the `struct` represents.

crates/modelardb_macros/src/lib.rs
skejserjensen: Ups, fixed.

crates/modelardb_macros/src/lib.rs
skejserjensen: And now that I have tried to do compile-time code generation with Zig's `comptime`, I have learned that it is apparently not possible [1,2]. Instead, one has to write their own functionality for code generation [3,4]. Luckily, that is fairly easy to do cmcuza/TerseTS#169, but it was very surprising to me how limited `comptime` is

1. - https://matklad.github.io/2025/04/19/things-zig-comptime-wont-do.html
2. - https://ziggit.dev/t/generating-tests-at-comptime/6473
3. - https://codeberg.org/ifreund/zig-wayland
4. - https://github.com/castholm/zigglgen/tree/master

Issue #417, opened 2026-08-12
Use modelardb_embedded Client in modelardb_client CLI


CGodiksen:
Since `Client` in `modelardb_embedded` now supports all the functionality of the Apache Arrow Flight server, we could consider using it in `modelardb_client` instead of the current custom client implementation used there. There is a slight dependency cost for the currently very thin CLI but refactoring it to use `Client` from `modelardb_embedded` would also make it possible to easily extend the functionality of the CLI.

We could also consider using `Client` in the integration test `TestContext` but this might not be ideal due to the current implementation of `TestContext` that allows for more raw configuration than what `Client` exposes. Wiring server integration tests through another crate might also be bad for the objective of the tests.

Issue #383, opened 2026-03-23
Add a more aggressive retry strategy for cluster operations and handle dead nodes


CGodiksen:
Currently we just return an error if any cluster operation fails on any node. Instead we should have a more aggressive retry strategy to ensure that cluster operations are forced through and if a node cannot be reached we should handle it.

PR #418, opened 2026-08-12
Add actions to list nodes and retrieve node metrics

CGodiksen:
This PR adds support for two new actions, `ListNodes` and `NodeMetrics`. `ListNodes` lists all the nodes in the cluster, or the node itself if it is a single node deployment. `NodeMetrics` returns a collection of metrics that show general system metrics such as CPU, disk, and memory usage, and ModelarDB specific metrics such as storage engine memory usage.

Methods have also been added to `Client` in `modelardb_embedded` to support node related operations on the client. Note that the client still implements the `Operations` trait for functionality that is shared between `DataFolder` and `Client` so these new methods are only extra functionality that only a `Client` supports such as `FlushNode`. Someone using the Rust API for `Client` now has access to all Apache Arrow Flight server functionality through a single API, instead of having to use `Client` for some things and a manual Apache Arrow Flight client for other things.

After the new `Client` was implemented, it was considered if it could be used elsewhere in the system since it now supports all the Apache Arrow Flight functionality. An issue (https://github.com/ModelarData/ModelarDB-RS/issues/417) has been created to explore this further.

### DIFF
--- a/crates/modelardb_embedded/src/operations/client.rs
+++ b/crates/modelardb_embedded/src/operations/client.rs
@@ -146,10 +146,11 @@ impl Client {
     async fn retrieve_action_bytes(&mut self, action_type: &str) -> Result<Bytes> {
         let mut response = self.send_action(action_type, vec![]).await?;
 
-        let message = response
-            .message()
-            .await?
-            .expect("Flight message should exist.");
+        let message = response.message().await?.ok_or_else(|| {
+            ModelarDbEmbeddedError::from(Status::internal(format!(
+                "Action '{action_type}' did not return a response message."
+            )))
+        })?;
 
         Ok(message.body)
     }
--- a/crates/modelardb_embedded/src/operations/client.rs
+++ b/crates/modelardb_embedded/src/operations/client.rs
@@ -170,6 +170,32 @@ impl Client {
 
         Ok(response.into_inner())
     }
+
+    /// Returns the URL of the cloud node that the node assigns to execute the SQL in `sql`. If the
+    /// node is not running in a cluster, or a cloud node could not be assigned,
+    /// [`ModelarDbEmbeddedError`] is returned.
+    pub async fn cloud_query_node(&mut self, sql: &str) -> Result<String> {
+        let flight_descriptor = FlightDescriptor::new_cmd(sql.to_owned());
+        let flight_info = self
+            .flight_client
+            .get_flight_info(Request::new(flight_descriptor))
+            .await?
+            .into_inner();
+
+        let endpoint = flight_info.endpoint.into_iter().next().ok_or_else(|| {
+            ModelarDbEmbeddedError::InvalidArgument(
+                "The node did not return an endpoint for the query.".to_owned(),
+            )
+        })?;
+
+        let location = endpoint.location.into_iter().next().ok_or_else(|| {
+            ModelarDbEmbeddedError::InvalidArgument(
+                "The endpoint did not return a cloud node location for the query.".to_owned(),
+            )
+        })?;
+
+        Ok(location.uri)
+    }
 }
 
 #[async_trait]
--- a/crates/modelardb_embedded/src/operations/client.rs
+++ b/crates/modelardb_embedded/src/operations/client.rs
@@ -28,7 +28,7 @@ use arrow::record_batch::RecordBatch;
 use arrow_flight::decode::FlightRecordBatchStream;
 use arrow_flight::encode::FlightDataEncoderBuilder;
 use arrow_flight::flight_service_client::FlightServiceClient;
-use arrow_flight::{Action, Criteria, FlightDescriptor, Ticket};
+use arrow_flight::{Action, Criteria, FlightDescriptor, Result as FlightResult, Ticket};
 use async_trait::async_trait;
 use datafusion::error::DataFusionError;
 use datafusion::execution::RecordBatchStream;
@@ -38,7 +38,7 @@ use modelardb_auth::BearerInterceptor;
 use prost::bytes::Bytes;
 use tonic::codegen::InterceptedService;
 use tonic::transport::{Channel, Endpoint};
-use tonic::{Request, Status};
+use tonic::{Request, Status, Streaming};
 
 use crate::error::{ModelarDbEmbeddedError, Result};
 use crate::operations::{
@@ -67,32 +67,30 @@ impl Client {
         Ok(Client { flight_client })
     }
 
-    /// Send the action with the type `action_type` and an empty body to the node. If the action
-    /// could not be performed, [`ModelarDbEmbeddedError`] is returned.
-    async fn run_action(&mut self, action_type: &str) -> Result<()> {
+    /// Send the action with the type `action_type` and `body` to the node and return the response
+    /// stream. If the action could not be performed, [`ModelarDbEmbeddedError`] is returned.
+    async fn send_action(
+        &mut self,
+        action_type: &str,
+        body: Vec<u8>,
+    ) -> Result<Streaming<FlightResult>> {
         let action = Action {
             r#type: action_type.to_owned(),
-            body: vec![].into(),
+            body: body.into(),
         };
 
-        self.flight_client.do_action(Request::new(action)).await?;
+        let response = self.flight_client.do_action(Request::new(action)).await?;
 
-        Ok(())
+        Ok(response.into_inner())
     }
 
     /// Send the action with the type `action_type` and an empty body to the node and return the
     /// body of the response. If the action could not be performed, [`ModelarDbEmbeddedError`] is
     /// returned.
     async fn retrieve_action_bytes(&mut self, action_type: &str) -> Result<Bytes> {
-        let action = Action {
-            r#type: action_type.to_owned(),
-            body: vec![].into(),
-        };
-
-        let response = self.flight_client.do_action(Request::new(action)).await?;
+        let mut response = self.send_action(action_type, vec![]).await?;
 
         let message = response
-            .into_inner()
             .message()
             .await?
             .expect("Flight message should exist.");
@@ -110,21 +108,9 @@ impl Operations for Client {
 
     /// Returns the type of the ModelarDB node that the client is connected to.
     async fn modelardb_type(&mut self) -> Result<ModelarDBType> {
-        // Retrieve the node type from the ModelarDB node.
-        let action = Action {
-            r#type: "NodeType".to_owned(),
-            body: vec![].into(),
-        };
+        let bytes = self.retrieve_action_bytes("NodeType").await?;
 
-        let response = self.flight_client.do_action(Request::new(action)).await?;
-
-        let message = response
-            .into_inner()
-            .message()
-            .await?
-            .expect("Flight message should exist.");
-
-        ModelarDBType::from_str(str::from_utf8(&message.body)?)
+        ModelarDBType::from_str(str::from_utf8(&bytes)?)
     }
 
     /// Creates a table with the name in `table_name` and the information in `table_type`. If the
@@ -152,12 +138,7 @@ impl Operations for Client {
             }
         };
 
-        let action = Action {
-            r#type: "CreateTable".to_owned(),
-            body: protobuf_bytes.into(),
-        };
-
-        self.flight_client.do_action(action).await?;
+        self.send_action("CreateTable", protobuf_bytes).await?;
 
         Ok(())
     }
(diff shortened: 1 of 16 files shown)


crates/modelardb_embedded/src/operations/client.rs
```diff
@@ -65,6 +69,134 @@ impl Client {
...
+
+        Ok(message.body)
+    }
+
+    /// Send the action with the type `action_type` and `body` to the node and return the response
+    /// stream. If the action could not be performed, [`ModelarDbEmbeddedError`] is returned.
+    async fn send_action(
+        &mut self,
+        action_type: &str,
+        body: Vec<u8>,
```
skejserjensen, line 163: Maybe make `body` parameter `&[u8]` instead of requiring the caller to create a `Vec` to pass a body. Although I do not know if it make the conversion into the right format for `Action` more complex and thus make the code less readable.

crates/modelardb_embedded/src/operations/client.rs
```diff
@@ -65,6 +69,134 @@ impl Client {
...
+            r#type: action_type.to_owned(),
+            body: body.into(),
+        };
+
+        let response = self.flight_client.do_action(Request::new(action)).await?;
+
+        Ok(response.into_inner())
+    }
+
+    /// Returns the URL of the cloud node that the node assigns to execute the SQL in `sql`. If the
```
skejserjensen, line 175: Is the second node mentioned in the comment not also a cloud node, since we no longer have a manager? If that is the case, should the node not just always return itself if it is not running in a cluster configuration?

crates/modelardb_server/src/cluster.rs
```diff
@@ -36,6 +36,26 @@ use tonic::transport::Endpoint;
 use crate::context::Context;
 use crate::error::{ModelarDbServerError, Result};
 
+/// The different possible modes that a ModelarDB server can be deployed in, assigned when the
+/// server is started.
+#[derive(Clone)]
+pub(crate) enum ClusterMode {
+    SingleNode(Node),
+    MultiNode(Box<Cluster>),
```
skejserjensen: Is `Cluster` in ` Box` because `clippy` warns about its size? Otherwise, I do not understand why the `Box`. It would probably be good to add a comment describing why `Box` is used.

crates/modelardb_server/src/cluster.rs
```diff
@@ -357,10 +377,19 @@ impl Cluster {
         Ok(url.to_owned())
     }
 
+    /// Return all nodes currently in the cluster. If the nodes could not be retrieved, return
+    /// [`ModelarDbServerError`].
+    pub(crate) async fn nodes(&self) -> Result<Vec<Node>> {
+        self.remote_data_folder
```
skejserjensen, line 383: Since we read the cluster nodes from the data folder, can we get into situations where we return nodes to the caller of `nodes()` that no longer exist? And if so, is this a problem?

crates/modelardb_server/src/context.rs
```diff
@@ -490,6 +493,87 @@ impl Context {
...
         Ok(schema)
     }
+
+    /// Collect the current resource usage metrics of the node, including CPU, memory, disk, and
+    /// storage engine memory usage.
+    pub(crate) async fn node_metrics(&self) -> protocol::NodeMetrics {
+        let mut system = System::new();
+
+        // Sample the CPU twice, separated by the minimum update interval, since a single refresh
+        // reads zero.
```
skejserjensen: Unclear what it meant by "...reads zero." as it does not look like `system.refresh_cpu_usage()` return a value that is used.

crates/modelardb_server/src/context.rs
```diff
@@ -490,6 +493,87 @@ impl Context {
...
+        let storage_engine = self.storage_engine.read().await;
+
+        let ingested_reserved_memory_in_bytes =
+            configuration_manager.ingested_reserved_memory_in_bytes();
+        let uncompressed_reserved_memory_in_bytes =
+            configuration_manager.uncompressed_reserved_memory_in_bytes();
+        let compressed_reserved_memory_in_bytes =
+            configuration_manager.compressed_reserved_memory_in_bytes();
+
+        let ingested_used_memory_in_bytes = (ingested_reserved_memory_in_bytes as i64
```
skejserjensen: Maybe add a comment describing what this calculation does as it is probably a bit hard to understand for anybody that does not know how the storage engine manages memory.

crates/modelardb_server/src/context.rs
```diff
@@ -490,6 +493,87 @@ impl Context {
...
+    /// of the local data folder path. If no disk matches, e.g., because the data folder is in
+    /// memory, the largest-capacity disk is used instead. If no disks are found, `(0, 0)` is
+    /// returned.
+    fn local_data_folder_disk_space(&self) -> (u64, u64) {
+        let disks = Disks::new_with_refreshed_list();
+        let location = self.data_folders.local_data_folder.location();
+
+        let maybe_disk = disks
+            .iter()
+            .filter(|disk| location.starts_with(&*disk.mount_point().to_string_lossy()))
```
skejserjensen: I think I understand the `filter()` as it seems to remove all of the disks that the data folder is not stored in, however, I dod not understand the purpose of the `max_by_key()` and `or_else()` calls.

skejserjensen: Why dereference the string and then borrow it right after? Why is this needed?

crates/modelardb_types/src/flight/mod.rs
```diff
@@ -215,6 +216,34 @@ fn decode_generated_column_expressions(
     Ok(expressions)
 }
 
+/// Encode `nodes` into a [`ClusterNodes`](protocol::ClusterNodes) protobuf message and serialize it.
+pub fn encode_and_serialize_cluster_nodes(nodes: Vec<Node>) -> Vec<u8> {
```
skejserjensen, line 220: If the function immediately converts the `Vec` to an iterator, why not just make the function take any type that `impl` iterator or into iterator, depending on what is required?

skejserjensen (changes requested):

crates/modelardb_embedded/src/operations/client.rs
CGodiksen: It does make the conversion slightly more complex and the callers of the function either pass an empty `Vec` or a freshly encoded `Vec<u8>` so there is no real creation cost. I have left it as it was.

crates/modelardb_embedded/src/operations/client.rs
CGodiksen: No the second node ("that the node assigns to execute") is the node the Client is connected to. We designed `get_flight_info` to be used to query the remote data folder using cloud nodes so this method just re-uses that design. 

If there are no cloud nodes or if we are not running in a cluster, `get_flight_info` returns an error.

crates/modelardb_server/src/cluster.rs
CGodiksen: It used to be a Clippy error since `SingleNode` had no data but now that we have added `Node` to `SingleNode`, it no longer seems to be an issue. I have removed the `Box`.

crates/modelardb_server/src/cluster.rs
CGodiksen: Since we currently rely on nodes removing themselves from the remote data folder when either stopped using ctrl+c or when `KillNode` is called, there could be issues if the node is stopped in another way.

However, this is not a problem introduced in this PR and it was already a problem when retrieving peer nodes and sending out a request to, for example, create a table on all peer nodes. 

This will be looked at further in this issue: https://github.com/ModelarData/ModelarDB-RS/issues/383

crates/modelardb_server/src/context.rs
CGodiksen: I have updated the comment to make it clear that it is `global_cpu_usage()` that returns 0 if we do not call `refresh_cpu_usage()` twice.

crates/modelardb_server/src/context.rs
CGodiksen: I have added a comment. While adding the comment I also found that the `max(0)` was placed slightly wrong since it is the remaining memory that can be negative, and that is what we want to clamp to 0. This is also reflected in the comment.

crates/modelardb_server/src/context.rs
CGodiksen: I have tried to make it more clear by making the doc comment more general and moving the implementation details down to the code. I have also extended the comment to make it clearer why we take the longest mount point.

CGodiksen: It was because of some weird type conversion due to `to_string_lossy()` returning `Cow<str>`. I have changed the code so we use `StdPath` instead which makes it easier to understand.

crates/modelardb_types/src/flight/mod.rs
CGodiksen: I have kept the function as it was since the current callers do not need a generalized function and to keep it consistent with `deserialize_and_extract_cluster_nodes` and with the other encoders and decodes that also take concrete types.

skejserjensen (approved):

crates/modelardb_embedded/src/operations/client.rs
```diff
@@ -65,6 +69,134 @@ impl Client {
...
+            new_value,
+        };
+
+        self.send_action("UpdateConfiguration", update_configuration.encode_to_vec())
+            .await?;
+
+        Ok(())
+    }
+
+    /// Flushes all data currently in memory to disk. If the data could not be flushed,
```
chrthomsen: tiny detail: "all data *currently* in memory here", but "all data in memory" on l. 107

crates/modelardb_embedded/src/operations/client.rs
```diff
@@ -65,6 +69,134 @@ impl Client {
...
+    }
+
+    /// Flushes all data to disk, transfers it to the remote object store, removes the node from the
+    /// cluster if necessary, and kills the node process. Since the process is killed, a
+    /// conventional response cannot be returned, so a dropped connection is not treated as an
+    /// error. If the data could not be flushed before the node was killed,
+    /// [`ModelarDbEmbeddedError`] is returned.
+    pub async fn kill_node(&mut self) -> Result<()> {
+        // The node exits while handling this action, so the response stream is dropped.
+        let _ = self.send_action("KillNode", vec![]).await;
```
chrthomsen, line 122: should this be `await?` to pass on an error in case the data could not be flushed?

crates/modelardb_embedded/src/operations/client.rs
```diff
@@ -65,6 +69,134 @@ impl Client {
...
+
+    /// Returns the current resource usage metrics of the node. If the metrics could not be
+    /// retrieved, [`ModelarDbEmbeddedError`] is returned.
+    pub async fn node_metrics(&mut self) -> Result<protocol::NodeMetrics> {
+        let bytes = self.retrieve_action_bytes("NodeMetrics").await?;
+
+        Ok(protocol::NodeMetrics::decode(bytes)?)
+    }
+
+    /// Send the action with the type `action_type` and an empty body to the node and return the
```
chrthomsen: Here "Send" and "return" but above and below "Return*s*" is used. Should this be "Sends ... and returns ..."?

crates/modelardb_embedded/src/operations/client.rs
```diff
@@ -65,6 +69,134 @@ impl Client {
...
+        let message = response.message().await?.ok_or_else(|| {
+            ModelarDbEmbeddedError::from(Status::internal(format!(
+                "Action '{action_type}' did not return a response message."
+            )))
+        })?;
+
+        Ok(message.body)
+    }
+
+    /// Send the action with the type `action_type` and `body` to the node and return the response
```
chrthomsen: "Sends ... and returns ..."?

chrthomsen (approved):

crates/modelardb_embedded/src/operations/client.rs
CGodiksen: Fixed.

crates/modelardb_embedded/src/operations/client.rs
CGodiksen: This would always return an error due to the node exiting while handling the action. We might be able to do something by looking at the specific error and I have added it to my notes to look into this further.

crates/modelardb_embedded/src/operations/client.rs
CGodiksen: Fixed.

crates/modelardb_embedded/src/operations/client.rs
CGodiksen: Fixed.

PR #419, opened 2026-08-13
Share one local and one remote DataFolder instance throughout the server

### LINKED ISSUE #416: Share one local and one remote DataFolder instance throughout the server
Currently, the established pattern used throughout the server, for example in the `ConfigurationManager`, `DataTransfer`, `CompressedDataManager`, `UncompressedDataManager`, and `DataOptimizer` components, is that we clone `local_data_folder` and `remote_data_folder` and pass the cloned instances to the various components. 

Instead `DataFolder` should own a single copy of its fields and then the entire `DataFolder` struct should be shared instead using an `Arc`. The result should be that there is a single local `DataFolder` instance and a single remote `DataFolder` instance. The same logic should be applied to handle the way we currently clone either the local or remote `DataFolder` for the `query_data_folder`, if possible.

Before this change is implemented, we should make sure that this does not cause any issues with either correctness or performance. The `DataFolder` fields are mostly static but the dynamic cache and dynamic session context could cause issues. We should also double check if this is an issue elsewhere in the server or in other crates.

CGodiksen:
This PR closes #416 by sharing a single local data folder and a single remote data folder throughout the server by wrapping them in an `Arc`. While looking into if there were any problems with sharing a single `DataFolder` instance, a small locking inefficiency was found when opening a `DeltaTable`. This is also fixed in this PR. Finally, the PR also fixes some naming inconsistencies, specifically regarding arguments being named `local_data_folder` while actually being a path or an object store.

### DIFF
--- a/crates/modelardb_storage/src/data_folder/mod.rs
+++ b/crates/modelardb_storage/src/data_folder/mod.rs
@@ -825,8 +825,8 @@ impl DataFolder {
     /// [`ModelarDbStorageError`] if a connection to the Delta Lake cannot be established or the
     /// table does not exist.
     async fn delta_table_from_path(&self, table_path: &str) -> Result<DeltaTable> {
-        // Clone the cached table out if possible and drop the DashMap guard before loading. load()
-        // is an async IO call, and get_mut() holds a write guard on the shard, so loading while
+        // Clone the cached table if possible and drop the DashMap guard before loading. load() is
+        // an async IO call, and get_mut() holds a write guard on the shard, so loading while
         // holding it would serialize access to every table in the same shard.
         let maybe_delta_table = self
             .delta_table_cache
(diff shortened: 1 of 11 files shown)


crates/modelardb_storage/src/data_folder/mod.rs
```diff
@@ -824,10 +825,17 @@ impl DataFolder {
     /// [`ModelarDbStorageError`] if a connection to the Delta Lake cannot be established or the
     /// table does not exist.
     async fn delta_table_from_path(&self, table_path: &str) -> Result<DeltaTable> {
-        // Use the cache if possible and load to get the latest table data.
-        if let Some(mut delta_table) = self.delta_table_cache.get_mut(table_path) {
+        // Clone the cached table out if possible and drop the DashMap guard before loading. load()
```
skejserjensen: I do not understand what "... clone out ... " means.

skejserjensen (approved):

chrthomsen (approved):

crates/modelardb_storage/src/data_folder/mod.rs
CGodiksen: Fixed.

PR #421, opened 2026-08-18
Use modelardb_embedded Client in CLI

### LINKED ISSUE #417: Use modelardb_embedded Client in modelardb_client CLI

CGodiksen:
This PR closes https://github.com/ModelarData/ModelarDB-RS/issues/417 by refactoring the CLI in `modelardb_client` to use `Client` from `modelardb_embedded` instead of a separate manual Apache Arrow Flight client. This unifies the way we access the server in the system and removes unnecessary code in the CLI. Other than refactoring the existing functionality, new commands have also been added to the CLI so it supports all the relevant methods that we expose through the `Client` API.

The PR also changes tab completion slightly so we now update the possible completions when we create or drop tables, and it updates the `KillNode` action to make it consistent with how we handle `ctrl+c` in the server. We now no longer try to transfer data when killing a node, and now only flush to disk. If a user wants to flush to the remote object store before killing a node, they can call `FlushNode` separately.

### DIFF
--- a/crates/modelardb_client/src/main.rs
+++ b/crates/modelardb_client/src/main.rs
@@ -218,7 +218,7 @@ async fn execute_command(client: &mut Client, command_and_arguments: &str) -> Re
             Ok(())
         }
         // Update a setting in the configuration of the node.
-        "\\s" => {
+        "\\sc" => {
             let name =
                 command_and_arguments
                     .next()
@@ -258,7 +258,7 @@ async fn execute_command(client: &mut Client, command_and_arguments: &str) -> Re
                  \\dc                           Print the configuration of the node.\n\
                  \\dn                           Print the nodes in the cluster.\n\
                  \\dm                           Print the resource usage metrics of the node.\n\
-                 \\s SETTING [VALUE]            Set SETTING to VALUE, or unset SETTING if VALUE is omitted.\n\
+                 \\sc SETTING [VALUE]           Set SETTING to VALUE, or unset SETTING if VALUE is omitted.\n\
                  \\f                            Flushes data in memory to disk.\n\
                  \\F                            Flushes data in memory and disk to the object store.\n\
                  \\h                            Print documentation for all supported commands.\n\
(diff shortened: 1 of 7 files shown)


crates/modelardb_client/src/main.rs
```diff
@@ -165,7 +129,15 @@ async fn execute_queries_from_a_repl(
     // Execute commands and queries and print the result.
     while let Ok(line) = editor.readline("ModelarDB> ") {
         editor.add_history_entry(line.as_str())?;
-        execute_and_print_command_or_query(&mut flight_service_client, &line).await
+        execute_and_print_command_or_query(&mut client, &line).await;
+
+        // Refresh the table names for tab-completion if a table may have been created or dropped.
+        let first_word = line.split_whitespace().next().unwrap_or("");
```
skejserjensen, line 135: Instead of having to search through the string for whitespace and split it for each query, maybe it would be simpler to check if the prefix of the string is `CREATE ` or `DROP ` with a space at the end.

skejserjensen (approved):

crates/modelardb_client/src/main.rs
```diff
@@ -239,32 +205,85 @@ async fn execute_command(
...
+                 \\dt                            Print the name of all the tables.\n\
+                 \\dc                            Print the configuration of the node.\n\
+                 \\dn                            Print the nodes in the cluster.\n\
+                 \\dm                            Print the resource usage metrics of the node.\n\
+                 \\sc SETTING [VALUE]            Set SETTING to VALUE, or unset SETTING if VALUE is omitted.\n\
+                 \\f                             Flushes data in memory to disk.\n\
+                 \\F                             Flushes data in memory and disk to the object store.\n\
+                 \\h                             Print documentation for all supported commands.\n\
+                 \\k                             Flush data to disk, kill the node, and quit modelardb.\n\
+                 \\q                             Quit modelardb."
```
chrthomsen:
```suggestion
                 \\q                             Quit ModelarDB."
```

chrthomsen (approved):

crates/modelardb_client/src/main.rs
CGodiksen: I have kept it as it was since `split_whitespace()` is lazy.

PR #425, opened 2026-08-22
Update ast-grep rules to use the latest version

### LINKED ISSUE #192: Update missing_puncation.yml for latest ast-grep
The version of `ast-grep` to use in GitHub Actions was set to version 0.21.4 in #175 as version 0.22.0 breaks `missing_puncation.yml`. Thus, the rule should be updated if `ast-grep` was purposely changed,  if `missing_puncation.yml` broke due to a bug ` in `ast-grep` this should probably be reported as an issue.

skejserjensen:
Update the `ast-grep` rules so they work with the latest version of `ast-grep` to fix #192. The `ast-grep` GitHub Action has also been updated to always use the latest version of `ast-grep`.

### DIFF
--- a/.github/workflows/ast-grep-scan.yml
+++ b/.github/workflows/ast-grep-scan.yml
@@ -30,4 +30,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Ast-grep Scan
-        uses: ast-grep/action@latest
+	uses: ast-grep/action@v1.5.0
--- a/.github/workflows/ast-grep-scan.yml
+++ b/.github/workflows/ast-grep-scan.yml
@@ -30,6 +30,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Ast-grep Scan
-        uses: ast-grep/action@v1.5
-        with:
-          version: 0.21.4  # Version 0.22.0 breaks the missing_punctuation rule.
+        uses: ast-grep/action@latest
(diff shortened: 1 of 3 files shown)


.github/workflows/ast-grep-scan.yml
```diff
@@ -30,6 +30,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Ast-grep Scan
-        uses: ast-grep/action@v1.5
-        with:
-          version: 0.21.4  # Version 0.22.0 breaks the missing_punctuation rule.
+        uses: ast-grep/action@latest
```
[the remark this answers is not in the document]
skejserjensen: Fixed.

.github/workflows/ast-grep-scan.yml
```diff
@@ -30,6 +30,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Ast-grep Scan
-        uses: ast-grep/action@v1.5
-        with:
-          version: 0.21.4  # Version 0.22.0 breaks the missing_punctuation rule.
+	uses: ast-grep/action@v1.5.0
```
CGodiksen: is the indentation right here?

CGodiksen (approved):

.github/workflows/ast-grep-scan.yml
skejserjensen: Fixed.

chrthomsen (approved):

PR #426, opened 2026-08-24
Use Rust 1.97.1 to compile on Windows

skejserjensen:
This PR specifies a specific Rust version to use to avoid Rust 1.98.0, as it segfaults on Microsoft Windows when running our tests. @CGodiksen verified that it also happens with `main` and not just #424 and #425, so it seems to be a problem with Rust 1.98.0. I have not verified what causes this segfault, but it may be this issue rust-lang/rust#161441 as it was introduced in Rust 1.98.0 and should be a near guaranteed segfault according to the comments.

### DIFF
--- a/.github/workflows/build-lint-test-and-upload.yml
+++ b/.github/workflows/build-lint-test-and-upload.yml
@@ -78,8 +78,6 @@ jobs:
         az storage container create -n modelardb --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;QueueEndpoint=http://127.0.0.1:10001/devstoreaccount1;'
 
     # Update Rust.
-    - name: Rustup Update
-      run: rustup update
     - name: Rustup Add llvm-tools
       run: rustup component add llvm-tools-preview
     - name: Cargo Install grcov
--- a/.github/workflows/build-lint-test-and-upload.yml
+++ b/.github/workflows/build-lint-test-and-upload.yml
@@ -77,9 +77,7 @@ jobs:
         sleep 5 # Ensure azurite has fully finished creating the endpoint.
         az storage container create -n modelardb --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;QueueEndpoint=http://127.0.0.1:10001/devstoreaccount1;'
 
-    # Update Rust.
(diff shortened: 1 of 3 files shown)


docs/user/README.md
```diff
@@ -37,7 +37,7 @@ The following commands are for Ubuntu Server. However, equivalent commands shoul
 
 ### All
 
-2. Install the latest stable [Rust Toolchain](https://rustup.rs/).
+2. Install Rust 1.97.1 [Rust Toolchain](https://rustup.rs/).
```
chrthomsen: "Install version 1.97.1 of the [Rust Toolchain](https://rustup.rs)"?

chrthomsen (approved):

docs/user/README.md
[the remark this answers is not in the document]
skejserjensen: Fixed.

skejserjensen: Fixed.

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.97.1"
+components = ["clippy"]
```
CGodiksen: Can we include "llvm-tools-preview" here instead of having `rustup component add llvm-tools-preview` in the CI pipeline?

.github/workflows/build-lint-test-and-upload.yml
CGodiksen: Should this be removed as well so we upload binaries with the same Rust toolchain as is tested?

.github/workflows/build-lint-test-and-upload.yml
```diff
@@ -78,8 +78,6 @@ jobs:
         az storage container create -n modelardb --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;QueueEndpoint=http://127.0.0.1:10001/devstoreaccount1;'
 
     # Update Rust.
```
CGodiksen: Maybe change this to something related to installing cargo tools if both `rustup` tasks are removed since it is no longer Rust being updated.

CGodiksen (changes requested):

workingjubilee:
You can use https://github.com/rust-lang/cargo-bisect-rustc to attempt to identify the nightly that caused the issue. It does not require detailed knowledge of what is going wrong, only a test script that can reproduce the issue.

.github/workflows/build-lint-test-and-upload.yml
skejserjensen: Fixed.

.github/workflows/build-lint-test-and-upload.yml
skejserjensen: Fixed.

rust-toolchain.toml
skejserjensen: Fixed.

skejserjensen:
> You can use https://github.com/rust-lang/cargo-bisect-rustc to attempt to identify the nightly that caused the issue. It does not require detailed knowledge of what is going wrong, only a test script that can reproduce the issue.

Thank you for the information. I will try to set up a machine with Microsoft Windows and try running our tests with the nightlies between Rust 1.97.1 and Rust 1.98.0 to determine when it starts failing.

CGodiksen (approved):

PR #428, opened 2026-09-04
Switch allocator to jemalloc

skejserjensen:
This PR changes the allocator from `snmalloc` to `jemalloc` to fix #426 without requiring Rust to be fixed at version 1.97.1. A library allocator is used to ensure the same allocator is used for all binaries and platforms. `jemalloc` is used because it generally performs as well as the other allocators, is mature, is proven, is easy to tune through `MALLOC_CONF` and `/etc/malloc.conf`, and provides features for profiling memory allocation. The `tikv-jemallocator` crate is used as it is the most commonly recommended and downloaded `jemalloc` Rust wrapper.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -389,9 +389,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.43"
+version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


skejserjensen:
Closed as `jemalloc` support for `x86_64-pc-windows-msvc` is untested and `tikv-jemalloc-sys` did not build, which is a blocking problem as ModelarDB supports Microsoft Windows.

